PR #1, opened 2023-03-27
chore: add basic CI tests

fuchsnj:
- adds some basic CI tests (more will come later)
- fixed some failing tests since `Cargo.lock` was removed (since this is a library)
   - zstd tests asserts exact binary output, which changed with a version bump (tests changed to assert it can still decode)
   - deprecations are allowed, since minor version bumps can fail CI. A different approach may be needed here in the future.

### DIFF
--- a/lib/stdlib/src/encode_zstd.rs
+++ b/lib/stdlib/src/encode_zstd.rs
@@ -88,34 +88,3 @@ impl FunctionExpression for EncodeZstdFn {
         TypeDef::bytes().infallible()
     }
 }
-
-#[cfg(test)]
-mod test {
-    use super::*;
-    use base64::Engine;
-
-    fn decode_base64(text: &str) -> Vec<u8> {
-        let engine = base64::engine::GeneralPurpose::new(
-            &base64::alphabet::STANDARD,
-            base64::engine::general_purpose::GeneralPurposeConfig::new(),
-        );
-
-        engine.decode(text).expect("Cannot decode from Base64")
-    }
-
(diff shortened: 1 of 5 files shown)


lib/stdlib/src/encode_zstd.rs
```diff
@@ -88,34 +88,3 @@ impl FunctionExpression for EncodeZstdFn {
...
-
-        engine.decode(text).expect("Cannot decode from Base64")
-    }
-
-    test_function![
-        encode_zstd => EncodeZstd;
-
-        with_defaults {
-            args: func_args![value: value!("you_have_successfully_decoded_me.congratulations.you_are_breathtaking.")],
-            want: Ok(value!(decode_base64("KLUv/QBY/QEAYsQOFKClbQBedqXsb96EWDax/f/F/z+gNU4ZTInaUeAj82KqPFjUzKqhcfDqAIsLvAsnY1bI/N2mHzDixRQA").as_bytes())),
```
bruceg, line 111: Could this maybe be changed to assert things like:
* output size is different (smaller) than input
* output has zstd header
* output can be decoded back to the input with `decode_zstd`

lib/stdlib/src/lib.rs
```diff
@@ -9,6 +9,7 @@
     unused_comparisons
 )]
 #![allow(
+    deprecated,
```
bruceg, line 12: Are there things that this breaks now?

bruceg (approved):

PR #2, opened 2023-03-27
chore: add clippy check

fuchsnj:
Adds clippy to CI


.github/workflows/test.yml
```diff
@@ -36,3 +36,5 @@ jobs:
         run: |
           cd lib/tests
           cargo run
+      - name: Clippy
+        run: cargo clippy --all
```
bruceg, line 40: Does this need `--all-targets` too?

bruceg (approved):

PR #161, opened 2023-03-28
chore(compiler): Add accessor to resolve optionally-constant expressions

bruceg:
This is a bit of an RFC to see if this approach makes sense. I saw this potential opportunity while looking at something else. This accessor would enable writing functions that could compile down to a fixed path if the parameter was a literal. For example, the `hmac` function takes an algorithm, and if that name is known at compile time we can skip looking up the name for each event processed.

### DIFF
--- a/lib/compiler/src/function.rs
+++ b/lib/compiler/src/function.rs
@@ -220,24 +220,13 @@ impl ArgumentList {
         Box::new(self.required_expr(keyword)) as _
     }
 
-    #[cfg(feature = "expr-literal")]
     pub fn optional_literal(
         &self,
         keyword: &'static str,
     ) -> Result<Option<crate::expression::Literal>, Error> {
-        self.optional_expr(keyword)
+        self.optional_resolve_literal(keyword)
             .map(|expr| match expr {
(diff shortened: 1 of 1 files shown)


fuchsnj:
How is this different from the existing `Expression::as_value()`, or were you not aware that existed?

The implementation is currently very naive and doesn't pick up things like literal assignments to variables, but we can certainly expand on that as needed.

bruceg:
> How is this different from the existing `Expression::as_value()`?

And this would be why we don't let me hack on VRL without training wheels.

I guess it only differs in that it resolves literals from variables as well. Could `as_value` be expanded to do that without also passing a context parameter?

fuchsnj:
> I guess it only differs in that it resolves literals from variables as well. Could as_value be expanded to do that without also passing a context parameter?

Yes, this is definitely something we want but haven't prioritized. There's actually a PR from awhile ago with the initial implementation that was abandoned: https://github.com/vectordotdev/vector/pull/10340

(I also think `as_value` needs renamed to help increase visibility, maybe something like `resolve_constant`?)

bruceg:
Closing as pointless

PR #167, opened 2023-03-29
chore: rename `Expression:as_value()` to `Expression::resolve_constant()`

fuchsnj:
I don't think it was clear what the `as_value` function really did. Renaming it should help.


spencergilbert (approved):
Agree with the rename being more clear, and assuming the compiler would have caught anywhere we didn't rename this... 👍

PR #172, opened 2023-03-31
chore: add wasm stub for `get_hostname`

fuchsnj:
`get_hostname` is not supported in WASM, but it's useful to include the function and abort when it's used instead. This prevents users from being confused when they can't find the function, and still allows checking if a VRL program compiles.

This also adds a CI check to ensure WASM can compile.


.github/workflows/test.yml
```diff
@@ -32,9 +32,16 @@ jobs:
...
+          rustup target add wasm32-unknown-unknown
+          cargo check --target wasm32-unknown-unknown
       - name: Run VRL tests
         run: |
           cd lib/tests
           cargo run
       - name: Clippy
         run: cargo clippy --workspace --all-targets -- -D warnings
+
+
```
spencergilbert, line 47:
```suggestion
```
?

lib/stdlib/Cargo.toml
```diff
@@ -29,7 +30,7 @@ flate2 = { version = "1.0.25", default-features = false, features = ["default"],
 grok = { version = "2", optional = true }
 hex = { version = "0.4", optional = true }
 hmac = { version = "0.12.1", optional = true }
-hostname = { version = "0.3", optional = true }
+#hostname = { version = "0.3", optional = true }
```
spencergilbert, line 33:
```suggestion
```

spencergilbert (approved):

lib/stdlib/src/wasm_unsupported_function.rs
```diff
@@ -0,0 +1,28 @@
...
+    pub fn new(span: Span, type_def: TypeDef) -> Self {
+        Self { span, type_def }
+    }
+}
+
+impl FunctionExpression for WasmUnsupportedFunction {
+    fn resolve(&self, _: &mut Context) -> Resolved {
+        Err(ExpressionError::Abort {
+            span: self.span,
+            message: Some("This function is not supported in web-assembly".to_owned()),
```
spencergilbert, line 21:
```suggestion
            message: Some("This function is not supported in WebAssembly".to_owned()),
```
Technically?

jszwedko (commented):
Not a big deal, but this function seems like one where we could return a dummy value like "playground.vrl.dev".

fuchsnj:
> Not a big deal, but this function seems like one where we could return a dummy value like "playground.vrl.dev".

I thought about that while writing this. I might follow up with a PR to add a dummy value so the rest of the helper code here can be merged in / used for the others I'm about to add. We probably want something more generic though since this will be used for _any_ usage of WASM, not just the playground.

spencergilbert:
`wasm.your.browser`

jszwedko:
> > Not a big deal, but this function seems like one where we could return a dummy value like "playground.vrl.dev".
> 
> I thought about that while writing this. I might follow up with a PR to add a dummy value so the rest of the helper code here can be merged in / used for the others I'm about to add. We probably want something more generic though since this will be used for _any_ usage of WASM, not just the playground.

👍 makes sense since we'll need the rest of the code here for functions that really can't be supported at the moment (like `parse_groks`).

PR #179, opened 2023-03-31
split test suite into additional jobs

spencergilbert:


.github/workflows/test.yml
```diff
@@ -24,22 +24,41 @@ env:
...
+        run: cargo fmt --check --all
+
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: "test"
+        run: cargo test --workspace
+
+  vrl:
```
spencergilbert, line 48: Don't love this name... but don't know if I have better suggestion.

fuchsnj: seems fine to me

fuchsnj (approved):

PR #181, opened 2023-03-31
add cache for ci

spencergilbert:


.github/workflows/test.yml
```diff
@@ -28,27 +28,43 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
```
fuchsnj, line 32: Why is this needed if rust is pre-installed + we use a `rust-toolchain.toml` (which might not always be the latest stable)

spencergilbert:
```
# selecting a toolchain either by action or manual `rustup` calls should happen
# before the plugin, as the cache uses the current rustc version as its cache key
```

I thought it'd be cleaner with the action vs shelling out, but it's also breaking due to the toolchain file.

spencergilbert: Reverted in [e3b3198](https://github.com/vectordotdev/vrl/pull/181/commits/e3b31989f9f997dedfed87207b180fc44f7f57a0)

fuchsnj (approved):

PR #193, opened 2023-04-16
feat(stdlib): Add `ingress_upstreaminfo` log format to `parse_nginx_log` function

### LINKED ISSUE #5: VRL: Add ingress-nginx log format for `parse_nginx_log`
### A note for the community

### Use Cases

Hey guys,

Currently there are only two log formats supported by `parse_nginx_log` function: "combined" and "error".
There's one more ingress-nginx controller default log format described [here](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/log-format/).

```
log_format upstreaminfo
    '$remote_addr - $remote_user [$time_local] "$request" '
    '$status $body_bytes_sent "$http_referer" "$http_user_agent" '
    '$request_length $request_time [$proxy_upstream_name] [$proxy_alternative_upstream_name] $upstream_addr '
    '$upstream_response_length $upstream_response_time $upstream_status $req_id';
```

While the beginning of the format is pretty much the same as "combined" described [here](https://nginx.org/en/docs/http/ngx_http_log_module.html#log_format):

```
log_format combined '$remote_addr - $remote_user [$time_local] '
                    '"$request" $status $body_bytes_sent '
                    '"$http_referer" "$http_user_agent"';
```
I believe there is an intention to make different default formats to parse them separately, like for example

```
structured, err = parse_nginx_log(.message, "ingress_upstreaminfo")
```

as discussed briefly in this [issue](https://github.com/vectordotdev/vector/issues/6103#issuecomment-762893997)

Anyways, in general it would be much easier for all kubernetes users who uses ingress-nginx in clusters to use `parse_nginx_log` instead of making their own regular expressions for a pretty default log format

### Attempted Solutions

_No response_

### Proposal

_No response_

### References

_No response_

### Version

_No response_

esergion:
Fixes #5 

Also, what do you guys think on renaming parsed fields like they are named in Nginx docs [here](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/log-format/)?

I would stick with that nginx's internal fields naming "convention", it's so close to RFCs at least.

If you are agree with this, I would make another PR to rename fields in "combined" format too


lib/stdlib/src/parse_nginx_log.rs
```diff
@@ -168,6 +175,34 @@ fn kind_combined() -> BTreeMap<Field, Kind> {
...
+        ("status".into(), Kind::integer()),
+        ("body_bytes_size".into(), Kind::integer()),
+        ("http_referer".into(), Kind::bytes()),
+        ("http_user_agent".into(), Kind::bytes()),
+        ("request_length".into(), Kind::integer()),
+        ("request_time".into(), Kind::float()),
+        ("proxy_upstream_name".into(), Kind::bytes()),
+        (
+            "proxy_alternative_upstream_name".into(),
+            Kind::bytes().or_null(),
```
fuchsnj, line 196:
```suggestion
            Kind::bytes().or_undefined(),
```

lib/stdlib/src/parse_nginx_log.rs
```diff
@@ -168,6 +175,34 @@ fn kind_combined() -> BTreeMap<Field, Kind> {
     ])
 }
 
+fn kind_ingress_upstreaminfo() -> BTreeMap<Field, Kind> {
+    BTreeMap::from([
+        ("remote_addr".into(), Kind::bytes()),
```
fuchsnj, line 180: This is optional.
```suggestion
        ("remote_addr".into(), Kind::bytes().or_undefined()),
```

lib/stdlib/src/parse_nginx_log.rs
```diff
@@ -168,6 +175,34 @@ fn kind_combined() -> BTreeMap<Field, Kind> {
...
+        ("remote_user".into(), Kind::bytes().or_null()),
```
fuchsnj, line 181: null means that the key actually exists, and has a null value. In this case, the remote_user key might not exist at all, so it should be undefined instead. Same for all optional fields here.
```suggestion
        ("remote_user".into(), Kind::bytes().or_undefined()),
```

lib/stdlib/src/parse_nginx_log.rs
```diff
@@ -168,6 +175,34 @@ fn kind_combined() -> BTreeMap<Field, Kind> {
...
+        ("timestamp".into(), Kind::timestamp()),
+        ("request".into(), Kind::bytes()),
+        ("method".into(), Kind::bytes()),
+        ("path".into(), Kind::bytes()),
+        ("protocol".into(), Kind::bytes()),
+        ("status".into(), Kind::integer()),
+        ("body_bytes_size".into(), Kind::integer()),
+        ("http_referer".into(), Kind::bytes()),
```
fuchsnj, line 189:
```suggestion
        ("http_referer".into(), Kind::bytes().or_undefined()),
```

lib/stdlib/src/parse_nginx_log.rs
```diff
@@ -168,6 +175,34 @@ fn kind_combined() -> BTreeMap<Field, Kind> {
...
+        ("http_user_agent".into(), Kind::bytes()),
```
fuchsnj, line 190:
```suggestion
        ("http_user_agent".into(), Kind::bytes().or_undefined()),
```

fuchsnj, line 190:
```suggestion
        ("http_user_agent".into(), Kind::bytes().or_undefined()),
```

fuchsnj (commented):
This mostly looks good. Just a few changes requested to the type definition.

lib/stdlib/src/parse_nginx_log.rs
```diff
@@ -107,6 +112,7 @@ fn regex_for_format(format: &[u8]) -> &Regex {
 fn time_format_for_format(format: &[u8]) -> String {
     match format {
         b"combined" => "%d/%b/%Y:%T %z".to_owned(),
+        b"ingress_upstreaminfo" => "%d/%b/%Y:%T %z".to_owned()
```
fuchsnj, line 115:
```suggestion
        b"ingress_upstreaminfo" => "%d/%b/%Y:%T %z".to_owned(),
```

fuchsnj:
> Also, what do you guys think on renaming parsed fields like they are named in Nginx docs [here](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/log-format/)?

That seems reasonable to me, although this would be a breaking change. Curious if anyone else has thoughts here. I think some field names were chosen initially to match Vector conventions (such as `timestamp`), but I think following those conventions will be less important in the future.

spencergilbert:
> That seems reasonable to me, although this would be a breaking change. Curious if anyone else has thoughts here. I think some field names were chosen initially to match Vector conventions (such as `timestamp`), but I think following those conventions will be less important in the future.

I'd agree with matching the "upstream" naming schemes 👍

esergion:
Thanks for the review, pushed all the requested changes

fuchsnj (approved):

PR #198, opened 2023-04-26
chore: migrate `Field` from the deprecated `FieldBuf` to `OwnedSegment`

fuchsnj:
The `Field` type used in `Collection` was migrated from the deprecated `FieldBuf` to `OwnedSegment`.

### DIFF
--- a/lib/lookup/src/lookup_v2/owned.rs
+++ b/lib/lookup/src/lookup_v2/owned.rs
@@ -262,6 +262,17 @@ impl From<OwnedValuePath> for String {
     }
 }
 
+impl Display for OwnedSegment {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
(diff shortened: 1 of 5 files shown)


lib/lookup/src/lookup_v2/owned.rs
```diff
@@ -262,6 +262,17 @@ impl From<OwnedValuePath> for String {
     }
 }
 
+impl Display for OwnedSegment {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        // This isn't the most efficient method, but it re-uses existing code
+        // Displaying is likely not in a hot path
```
spencergilbert, line 268: I don't love that its "likely", but I'm also failing to think of where it may be used that performance sensitive

spencergilbert (approved):

fuchsnj:
`OwnedSegment` was the wrong choice here... will reopen soon.

PR #200, opened 2023-04-26
chore: Migrate stdlib `get` function to new path code

fuchsnj:


spencergilbert (approved):
Looks equivalent to me

PR #203, opened 2023-04-26
chore: migrate path logic for `set` stdlib function

fuchsnj:


fuchsnj:
replaced by https://github.com/vectordotdev/vrl/pull/204

PR #212, opened 2023-05-01
chore: rename the `lookup` crate to `path`

fuchsnj:
Now that lookup V1 is gone, this re-organizes the crate a bit (removing the `lookup_v2` prefix) and renames the entire crate to `path` since the term "lookup" isn't used anymore.


lib/path/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![deny(warnings)]
```
spencergilbert, line 1: Are we working on moving this up to apply to everything in the project?

spencergilbert (approved):
Looks like a straight forward PR, left one question and wanted to make sure we don't need to make changes to Vector - this is all internal to VRL, not the public interface?

fuchsnj:
> wanted to make sure we don't need to make changes to Vector
This will definitely require changes in Vector. I have another PR coming out soon that will do some more re-organization of the crates, then I was planning to pull in those changes back into Vector.

lib/path/src/lib.rs
fuchsnj: I'm not really convinced either way yet on this. I think it makes sense on binaries where the dependencies are locked, but for a library, you can have dependency versions change at any time that could start to generate warnings and fail CI. I'm just keeping this as-is for now until it's more clear how to proceed here.

spencergilbert: Fair enough, I was definitely thinking of things from how we've been doing Vector

PR #213, opened 2023-05-01
chore: re-export all sub-crates in `vrl`

fuchsnj:
This re-exports all sub-crates in the root `vrl` crate, which enables using VRL as just a single crate.

In the future, more useful high-level abstractions / preludes may exist in the root `vrl` crate, but right now it's a fairly simple list of re-exports.


src/lib.rs
spencergilbert, line 1: Looking at the naming here, is there any reason we didn't just use the names not prefixed with `vrl_`? Say, rather than renaming `vrl_compiler` to `compiler` - I assume it's there historically, but we could probably drop that now that VRL is a separate project now?

spencergilbert (approved):

spencergilbert:
https://github.com/bruxisma/setup-cargo-hack for CI?

PR #219, opened 2023-05-04
added supported null \0 escape character to new lexer

arthmoeros:
added supported null \0 escape character to new lexer


StephenWakely (changes requested):
Nice. Could you please add a test to cover this. You should be able to do something very similar to this https://github.com/vectordotdev/vrl/blob/main/lib/parser/src/lex.rs#L1481-L1499

Let me know if you have any questions about the way that test is set up..

arthmoeros:
Hi @fuchsnj , I did miss @StephenWakely response, I'm sorry!

I was checking out the provided link, but I think it is broken (404). I was trying to find where to add a test for this, but I'm having trouble finding it.

fuchsnj:
> Hi @fuchsnj , I did miss @StephenWakely response, I'm sorry!
> 
> I was checking out the provided link, but I think it is broken (404). I was trying to find where to add a test for this, but I'm having trouble finding it.

There was some significant refactoring recently that moved a lot of files. The new link is:
https://github.com/vectordotdev/vrl/blob/main/src/parser/lex.rs#L1481-L1499

PR #220, opened 2023-05-05
chore: add `arbitrary` and `lua` features

fuchsnj:


spencergilbert (approved):
Is there a logic to how you've organized the features? Could we leave comments so it's easier to add/change them going forward?

PR #222, opened 2023-05-05
chore: Set up 3rd party license list file

bruceg:
This adds a generated `LICENSE-3rdparty.csv` file mandated by DataDog's open source policy, as well as a CI step to check that the file is up to date.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -67,6 +67,16 @@ jobs:
       - name: Check that all features can compile
         run: cargo hack check --feature-powerset --depth 1
 
+  check-licenses:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: Swatinem/rust-cache@v2
+      - name: Install the 3rd-party license tool
+        run: cargo install --git https://github.com/DataDog/rust-license-tool
+      - name: Check that the 3rd-party license file is up to date
+        run: rust-license-tool check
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -67,6 +67,15 @@ jobs:
       - name: Check that all features can compile
         run: cargo hack check --feature-powerset --depth 1
 
+  check-licenses:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: Swatinem/rust-cache@v2
+      - name: Check that the 3rd-party license file is up to date
+        run: cargo install --git https://github.com/DataDog/rust-license-tool
```
jszwedko, line 76: I assume this'll change when we have the first official release?

jszwedko (approved):
Fantastic! Nice work. It'd probably be good to get Nathan's review here too.

.github/workflows/test.yml
bruceg: What do you see as deficient with this approach? That is, change to what?

jszwedko: Ah, I just imagined we might want to version the license tool and, potentially, even publish it to install from crates.io.

bruceg: I guess, eventually. We're not there yet.

fuchsnj (approved):

PR #245, opened 2023-05-13
feat: add fuzzer

fuchsnj:
This adds the initial implementation of a fuzzer for VRL. It's still lacking a lot of features, but it's a start. See the README for more details.

You can see all the issues found with fuzz testing here: https://github.com/vectordotdev/vrl/issues?q=is%3Aissue+label%3Afuzz+


StephenWakely:
Nice. Is there a reason you chose `afl` over `libFuzzer`?

blt (approved):
Very cool. I'd have written this almost exactly the same way. You may want to increase the `input.txt` with more examples but that's the only note I have.

fuchsnj:
> You may want to increase the input.txt with more examples but that's the only note I have.

I plan on adding at least 1 input for each function, and possibly more after that. It's finding enough issues right now though, so I'm focusing on fixing issues first so it's not just finding a bunch of duplicates over and over.

PR #247, opened 2023-05-16
feat(stdlib): Add timezone argument to format_timestamp

nabokihms:
Related discussion
https://github.com/vectordotdev/vector/discussions/17374


src/stdlib/format_timestamp.rs
```diff
@@ -43,39 +53,66 @@ impl Function for FormatTimestamp {
...
+            None => format_timestamp_with_tz(bytes, ts, &Utc),
+            Some(tz) => {
+                let tz = &tz.resolve(ctx)?.try_bytes()?;
+                let tz = String::from_utf8_lossy(tz);
+                match tz {
+                    std::borrow::Cow::Borrowed("Local") => {
+                        format_timestamp_with_tz(bytes, ts, &Local)
+                    }
+                    _ => {
+                        let tz: Tz = tz.parse().unwrap();
```
jszwedko, line 99: Rather than `unwrap()` I think we'll want to return the error here. I think you should just be able to return a string like:

```suggestion
                        let tz: Tz = tz.parse().map_err(|err| format!("unable to parse timezone: {}", err))?;
```

src/stdlib/format_timestamp.rs
```diff
@@ -43,39 +53,66 @@ impl Function for FormatTimestamp {
...
 
-        format_timestamp(bytes, ts)
+        match self.tz.clone() {
+            None => format_timestamp_with_tz(bytes, ts, &Utc),
+            Some(tz) => {
+                let tz = &tz.resolve(ctx)?.try_bytes()?;
+                let tz = String::from_utf8_lossy(tz);
+                match tz {
+                    std::borrow::Cow::Borrowed("Local") => {
+                        format_timestamp_with_tz(bytes, ts, &Local)
```
jszwedko, line 96: For `Local` I think we'll actually want to use the `timezone` option on the `remap` transform if it has been specified. You can get that from the `ctx` object like you see in `parse_timestamp`:

https://github.com/vectordotdev/vrl/blob/43df3373c03125d53ee1638909227ce9eca66955/src/stdlib/parse_timestamp.rs#L8

jszwedko (commented):
Awesome, thanks for this @nabokihms ! I just left a couple of comments below.

src/stdlib/format_timestamp.rs
nabokihms: Took the `*ctx.timezone()`, but slightly different way than was suggested.

src/stdlib/format_timestamp.rs
nabokihms: Applied

src/stdlib/format_timestamp.rs
```diff
@@ -1,15 +1,20 @@
 use crate::compiler::prelude::*;
 use chrono::{
     format::{strftime::StrftimeItems, Item},
-    DateTime, Utc,
+    DateTime, TimeZone, Utc,
```
jszwedko, line 4: Actually, I realized we have our own `TimeZone` type that includes support for parsing `local` to indicate local time. Could we use that one instead?

https://github.com/vectordotdev/vrl/blob/04941e71487166c9746a504bf6fcddefd198b13d/src/compiler/datetime.rs#L15-L26

That should help ensure timezone handling is consistent within VRL.

src/stdlib/format_timestamp.rs
```diff
@@ -32,6 +37,11 @@ impl Function for FormatTimestamp {
                 kind: kind::BYTES,
                 required: true,
             },
+            Parameter {
+                keyword: "tz",
```
jszwedko, line 41: Somewhat of a nit, but I don't think we have a precedent for abbreviating parameter names. Could we call this `timezone`?

jszwedko (commented):
Thanks @nabokihms ! I left a couple of more comments. Apologies for not discovering them before.

src/stdlib/format_timestamp.rs
nabokihms: Sure, no problem.

src/stdlib/format_timestamp.rs
nabokihms: I refactored this part.

Another thing I found out is that I tried to parse the timezone during compilation, which needs to be corrected in terms of dynamically passed timezones (if the timezone is not constant but a variable).

The parsing part is moved to the function body. If my assumption is wrong, and we do not expect users to pass a variable as the timezone value, I can return the code back.

jszwedko:
Thanks @nabokihms ! I was going to push some other minor updates, but it doesn't look like I can push to this branch. This is the diff I have:

```patch
ba69f85f0 - (HEAD -> tz-format) style updates (60 seconds ago) <Jesse Szwedko>
diff --git a/CHANGELOG.md b/CHANGELOG.md
index 7c80ac172..3b41522eb 100644
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## unreleased
-- added the timezone (`tz`) argument to the `format_timestamp` vrl function.
+- added the `timezone` argument to the `format_timestamp` vrl function.
 
 ## `0.4.0` (2023-05-11)
 - consolidated all crates into the root `vrl` crate. The external API stayed the same, with the exception of macros, which are now all exported at the root of the `vrl` crate.
diff --git a/src/stdlib/format_timestamp.rs b/src/stdlib/format_timestamp.rs
index 6aa093260..c25068bc6 100644
--- a/src/stdlib/format_timestamp.rs
+++ b/src/stdlib/format_timestamp.rs
@@ -2,28 +2,19 @@ use crate::compiler::prelude::*;
 use crate::compiler::TimeZone;
 use chrono::{
     format::{strftime::StrftimeItems, Item},
-    DateTime, Local, Utc,
+    DateTime, Utc,
 };
-use std::borrow::Borrow;
 
 fn format_timestamp_with_tz(ts: Value, format: Value, timezone: Option<Value>) -> Resolved {
-    let format = format.try_bytes()?;
-    let format = String::from_utf8_lossy(&format);
-    let items = StrftimeItems::new(&format)
-        .map(|item| match item {
-            Item::Error => Err("invalid format".into()),
-            _ => Ok(item),
-        })
-        .collect::<ExpressionResult<Vec<_>>>()?;
-
     let ts: DateTime<Utc> = ts.try_timestamp()?;
 
-    let ts = match timezone {
-        None => ts.format_with_items(items.into_iter()).to_string(),
-        Some(timezone) => try_format_with_timezone(ts, items, timezone)?,
-    };
+    let format_bytes = format.try_bytes()?;
+    let format = String::from_utf8_lossy(&format_bytes);
 
-    Ok(ts).map(Into::into)
+    let timezone_bytes = timezone.map(|t| t.try_bytes()).transpose()?;
+    let timezone = timezone_bytes.as_ref().map(|b| String::from_utf8_lossy(&b));
+
+    try_format_with_timezone(ts, &format, timezone.as_deref()).map(Into::into)
 }
 
 #[derive(Clone, Copy, Debug)]
@@ -62,7 +53,7 @@ impl Function for FormatTimestamp {
     ) -> Compiled {
         let value = arguments.required("value");
         let format = arguments.required("format");
-        let timezone: Option<Box<dyn Expression>> = arguments.optional("timezone");
+        let timezone = arguments.optional("timezone");
 
         Ok(FormatTimestampFn {
             value,
@@ -99,14 +90,13 @@ impl FunctionExpression for FormatTimestampFn {
     fn resolve(&self, ctx: &mut Context) -> Resolved {
         let bytes = self.format.resolve(ctx)?;
         let ts = self.value.resolve(ctx)?;
+        let tz = self
+            .timezone
+            .as_ref()
+            .map(|tz| tz.resolve(ctx))
+            .transpose()?;
 
-        match self.timezone.clone() {
-            Some(tz) => {
-                let tz = tz.resolve(ctx)?;
-                format_timestamp_with_tz(ts, bytes, Some(tz))
-            }
-            None => format_timestamp_with_tz(ts, bytes, None),
-        }
+        format_timestamp_with_tz(ts, bytes, tz)
     }
 
     fn type_def(&self, _: &state::TypeState) -> TypeDef {
@@ -115,26 +105,33 @@ impl FunctionExpression for FormatTimestampFn {
 }
 
 fn try_format_with_timezone(
-    ts: DateTime<Utc>,
-    items: Vec<Item>,
-    timezone: Value,
+    dt: DateTime<Utc>,
+    format: &str,
+    timezone: Option<&str>,
 ) -> ExpressionResult<String> {
-    let timezone = timezone.try_bytes()?;
-    let timezone = String::from_utf8_lossy(&timezone);
-    let parsed_timezone = TimeZone::parse(timezone.borrow());
-
-    match parsed_timezone {
-        Some(parsed_timezone) => match parsed_timezone {
-            TimeZone::Local => Ok(ts
-                .with_timezone(&Local)
-                .format_with_items(items.into_iter())
-                .to_string()),
-            TimeZone::Named(tz) => Ok(ts
-                .with_timezone(&tz)
-                .format_with_items(items.into_iter())
-                .to_string()),
-        },
-        None => Err(format!("unable to parse timezone: {timezone}").into()),
+    let items = StrftimeItems::new(format)
+        .map(|item| match item {
+            Item::Error => Err("invalid format".into()),
+            _ => Ok(item),
+        })
+        .collect::<ExpressionResult<Vec<_>>>()?;
+
+    let timezone = timezone
+        .map(|timezone| {
+            TimeZone::parse(timezone).ok_or(format!("unable to parse timezone: {timezone}"))
+        })
+        .transpose()?;
+
+    match timezone {
+        Some(TimeZone::Named(tz)) => Ok(dt
+            .with_timezone(&tz)
+            .format_with_items(items.into_iter())
+            .to_string()),
+        Some(TimeZone::Local) => Ok(dt
+            .with_timezone(&chrono::Local)
+            .format_with_items(items.into_iter())
+            .to_string()),
+        None => Ok(dt.format_with_items(items.into_iter()).to_string()),
     }
 }
 ```

Any chance you could apply that here?

nabokihms:
Applied the patch with some minor additions to satisfy clippy.

jszwedko (approved):
This looks good to me, thanks @nabokihms !

Issue #238, opened 2023-05-12
Multiplication overflows can panic (in debug builds)


fuchsnj:
The following VRL code will panic (when compiled in debug mode). In release mode, it will wrap.

```coffee
444111111111*1111111111111111
```

VRL version: `0.4.0`

Issue #242, opened 2023-05-13
Assigning values to large array indices can use excessive cpu/memory


fuchsnj:
Assigning a value to a very large (positive or negative) index will expand the array to the size required to fit that index, potentially using an excessive amount of memory. A large limit (e.g. 1 million?) that causes an abort may help prevent abuse or accidents here.

Examples
```coffee
x[99999999999] = true;
```
```coffee
x[-99999999999] = true;
```

fuchsnj:
This would require https://github.com/vectordotdev/vrl/issues/244

PR #252, opened 2023-05-23
fix: arithmetic overflow panic

fuchsnj:
closes https://github.com/vectordotdev/vrl/issues/238

Multiplication, addition and subtraction overflows could panic when running in a debug build. This is fixed to always use wrapping arithmetic.


src/compiler/value/arithmetic.rs
```diff
@@ -73,7 +73,10 @@ impl VrlValueArithmetic for Value {
             Value::Integer(lhv) if rhs.is_float() => {
                 Value::from_f64_or_zero(lhv as f64 * rhs.try_float()?)
             }
-            Value::Integer(lhv) => (lhv * rhs.try_into_i64().map_err(|_| err())?).into(),
+            Value::Integer(lhv) => {
+                let rhv = rhs.try_into_i64().map_err(|_| err())?;
+                i64::wrapping_mul(lhv, rhv).into()
+            }
```
jszwedko, line 79: Is overflow the general pattern we use arithmetic? Just figured it's worth writing that in https://github.com/vectordotdev/vector/blob/master/docs/DEVELOPING.md (or adding a similar file to VRL).

fuchsnj: I'm open to suggestions, but I figured it's worth at least making `debug` builds match what `release` builds currently do. Returning errors would require all multiplication to be fallible, which could be pretty painful to use. Another option might be to convert the result to an `f64` if it would overflow, but again that would change the type definition and probably make it pretty painful to use in practice. I can add a line to `DEVELOPING.md`.

fuchsnj: Although you bring up a good point, this should probably be consistent across all arithmetic. I'll modify those too. (addition, subtraction, etc.)

src/compiler/value/arithmetic.rs
```diff
@@ -107,18 +112,20 @@ impl VrlValueArithmetic for Value {
...
             (Value::Float(lhs), rhs) => (lhs
                 + rhs
                     .try_into_f64()
                     .map_err(|_| ValueError::Add(Kind::float(), rhs.kind()))?)
             .into(),
             (lhs @ Value::Bytes(_), Value::Null) => lhs,
             (Value::Bytes(lhs), Value::Bytes(rhs)) => {
+                #[allow(clippy::integer_arithmetic)]
                 let mut value = BytesMut::with_capacity(lhs.len() + rhs.len());
```
jszwedko, line 129: I wonder if we wouldn't actually _want_ this to panic. The behavior otherwise seems like it'd be strange.

jszwedko (approved):
Great!

src/compiler/value/arithmetic.rs
fuchsnj: Today you could never get into a situation where this would overflow. If either of these values are large enough to cause an overflow, that means you would have already allocated several million terrabytes :)


In general though, I'd like to add some kind of allocation limit in VRL for https://github.com/vectordotdev/vrl/issues/242

Issue #239, opened 2023-05-12
Nested assignment in query can create an invalid type def


fuchsnj:
The following VRL code creates an invalid type definition

```coffee
$ type_def(.)
{ "object": {  }, "object_unknown_infinite": { "any": true } }

$ {. =null}.y
null

$ .
null

$ type_def(.)
{ "object": {  }, "object_unknown_infinite": { "any": true } }

```

VRL version: `0.4.0`

PR #258, opened 2023-05-25
fix: typedefs with query side-effects

fuchsnj:
closes https://github.com/vectordotdev/vrl/issues/239

queries in VRL can query both functions and "containers" (arbitrary code blocks), which can contain side-effects. These side-effects were not captured in the type definition.


jszwedko (approved):
Definitely not a VRL compiler expert but the changes generally make sense to me. The tests definitely do.

Issue #148, opened 2022-03-21
vrl: null and unicode escapes not working in string literals


hhromic:
### A note for the community

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Problem

I recently found out that the null and unicode character escape sequences are not working in [string literal expressions in VRL](https://vector.dev/docs/reference/vrl/expressions/#string).
The documentation makes sense, therefore I think something is off in the VRL parser.

Examples from the VRL REPL:
```
$ "hello\0world"
error[E202]: syntax error
  ┌─ :1:1
  │
1 │ "hello\0world"
  │ ^^^^^^^^^^^^^^ unexpected error: invalid escape character: \0
  │
  = see language documentation at https://vrl.dev
```
```
$ "hello\u{1F30E}world"  # also tested with \U{1F30E} and \u1F30E just in case
error[E202]: syntax error
  ┌─ :1:1
  │
1 │ "hello\u{1F30E}world"
  │ ^^^^^^^^^^^^^^^^^^^^^ unexpected error: invalid escape character: \u
  │
  = see language documentation at https://vrl.dev
```

All the other documented escape sequences seems to be working fine (although not all printed correctly):
```
$ "hello\nworld"
"hello\nworld"

$ "hello\"world"
"hello\"world"

$ "hello\'world"
"hello'world"

$ "hello\\world"
"hello\\world"

$ "hello\nworld"
"hello\nworld"

$ "hello\rworld"  # \r (carriage-return) was rendered instead of printed with an escape
world"

$ "hello\tworld"  # \t (tab) was rendered instead of printed with an escape
"hello  world"
```

### Configuration

```text
N/A
```

### Version

vector 0.20.0 (x86_64-unknown-linux-gnu 2a706a3 2022-02-11)

arthmoeros:
Is anyone by any chances working on this?, I did try to add null escape in the mentioned new lexer, but I can't get it to work, still says `\0` is an invalid escape.

I'm not very well versed on Rust so maybe I'm missing something, this is my fork, if anybody could point me in the right direction.

https://github.com/vectordotdev/vrl/compare/main...arthmoeros:vrl:null_escape_lexer

[...]

PR #259, opened 2023-05-26
Added null escape to lexer

arthmoeros:
New PR (previous one -> https://github.com/vectordotdev/vrl/pull/219), had to make a new one because of conflicts.

Partially solves https://github.com/vectordotdev/vrl/issues/148

added supported null \0 escape character to new lexer

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,7 +13,7 @@ provide everything you need to get started.
    GitHub account (only applicable to outside contributors).
 3. [Create a new Git branch][urls.create_branch].
 4. Make your changes.
-5. Add and/or update unit tests to cover your changes.
+5. Add and/or update tests to cover your changes.
 6. Add a one line summary of changes to the "unreleased" section of `CHANGELOG.md`.
 7. [Submit the branch as a pull request][urls.submit_pr] to the repo. A team member should
    comment and/or review your pull request.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,8 +13,9 @@ provide everything you need to get started.
    GitHub account (only applicable to outside contributors).
 3. [Create a new Git branch][urls.create_branch].
 4. Make your changes.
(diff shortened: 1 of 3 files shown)


src/parser/lex.rs
```diff
@@ -1308,6 +1308,7 @@ fn unescape_string_literal(mut s: &str) -> String {
                 b'n' => '\n',
                 b'r' => '\r',
                 b't' => '\t',
+                b'\0' => '\0',
```
fuchsnj, line 1311:
```suggestion
                b'0' => '\0',
```
The character after the backslash is a literal `0` not null. This results in a panic (since it reaches the `unimplemented` below). The test you added was fine, but it only tests the lexer. The panic occurs in the parser (which runs after the lexer). I would suggest also adding a unit test to the `unescape_string_literal` function for the null character, and maybe consider adding it to our VRL test suite, in  a new file in `lib/tests/tests/expressions/literal` which will run it through the entirety of VRL.

I also want to point out you can use the VRL CLI (by running `cargo run` in `lib/cli`) when testing locally. This lets you easily test out new features like this.

Let me know if you need any help adding more tests for this. After this change, it appears to work for me in the CLI.

fuchsnj (changes requested):

CONTRIBUTING.md
```diff
@@ -13,8 +13,9 @@ provide everything you need to get started.
    GitHub account (only applicable to outside contributors).
 3. [Create a new Git branch][urls.create_branch].
 4. Make your changes.
-5. Add a one line summary of changes to the "unreleased" section of `CHANGELOG.md`.
-6. [Submit the branch as a pull request][urls.submit_pr] to the repo. A team member should
+5. Add and/or update unit tests to cover your changes.
```
fuchsnj, line 16: Thanks for adding this. I suggest a minor edit here, since we have several different ways to test features. Not all of them are "unit" tests.
```suggestion
5. Add and/or update tests to cover your changes.
```

fuchsnj (approved):
This seems to be working. Thanks!

Issue #240, opened 2023-05-13
`Program` doesn't give the type definition of the program result


fuchsnj:
Currently the `Program` type (returned from compiling a VRL program) gives you access to the type definitions of the target and locals through the `Program::final_type_state` function. There is no type definition for the result itself (the `Value` returned from `resolve`).

PR #262, opened 2023-05-31
feat: give access to the program result type definition

fuchsnj:
closes: https://github.com/vectordotdev/vrl/issues/240

The type definition of both the result and target of VRL programs is known, but only the target was exposed to users. This gives access to both. It is a minor breaking change.

This is useful for things such as Vector VRL conditions to be able to verify at compile time that programs return a boolean. It will also help with fuzz testing.

### DIFF
--- a/src/compiler/expression/assignment.rs
+++ b/src/compiler/expression/assignment.rs
@@ -770,7 +770,7 @@ mod test {
         assert_eq!(
             result
                 .program
-                .final_type_state()
+                .final_type_info()
                 .state
                 .external
                 .target()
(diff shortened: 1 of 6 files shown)


src/compiler/expression/assignment.rs
```diff
@@ -771,6 +771,7 @@ mod test {
             result
                 .program
                 .final_type_state()
+                .state
```
spencergilbert, line 774: Calling `.state` on something called `_state()` feels off, any thoughts on a rename to make the result of `.final_type_state()` clearer? `.final_type_info()` matching the return maybe?

spencergilbert (approved):

PR #287, opened 2023-06-29
chore(ci): fix team membership action

dsmith3197:
`GH_PAT_ORG` has the correct permissions for `read:org`


dsmith3197:
@fuchsnj I'm not sure if `GH_PAT_ORG` is available in this repo so we may need to add it.

.github/workflows/gardener_open_pr.yml
```diff
@@ -18,7 +18,7 @@ jobs:
         with:
           username: ${{ github.actor }}
           team: vector
-          GITHUB_TOKEN: ${{ secrets.GH_PROJECT_PAT }}
+          GITHUB_TOKEN: ${{ secrets.GH_PAT_ORG }}
```
jszwedko, line 21: Hmm, I'm confused. Maybe the wrong token was added for `GH_PROJECT_PAT` for this repo. The `GH_PROJECT_PAT` token has these scopes:

```
❯  curl -sS -f -I -H "Authorization: token ${GITHUB_TOKEN}" https://api.github.com | grep ^x-oauth-scopes: | cut -d' ' -f2- | tr -d "[:space:]" | tr ',' '\n'
project
read:org
repo%
```

Let me replace the `GH_PROJECT_PAT`.

jszwedko: Replaced `GH_PROJECT_PAT`.

dsmith3197: @jszwedko Just to confirm, did you add the `read:org` scope to `GH_PROJECT_PAT`? If so, we can close this PR.

jszwedko: Seems like it still added my test PR: https://github.com/vectordotdev/vrl/actions/runs/5426976626/jobs/9869715707?pr=288

Maybe it needs additional scopes?

jszwedko: It already had the `read:org` scope 🤔

dsmith3197: Interesting, let's wait until @neuronull is back, as he created the other token that has the correct permissions.

neuronull: I believe the only permission that `GH_PAT_ORG` has, is `read:org` .

Perhaps it's not an issue of whether `GH_PROJECT_PAT` _needs_ more permissions, but an issue that it _has_ more? The code for `get-user-teams-membership` doesn't do explicit checks AFAICT, so if that were the case, it would seem to be part of the includes there.

Also, if we accept this change to use `GH_PAT_ORG` , we'll have to add it to this repo for it to work. Hmmm it appears I may not have recorded the token for `GH_PAT_ORG` ... shoot. If we want to use one token, might have to generate a new one.

dsmith3197 (commented):
@jszwedko I can't approve my own PR but your changes look good to me!

neuronull (approved):

.github/workflows/gardener_open_pr.yml
jszwedko: Ended up swapping out with the token from https://github.com/vectordotdev/vector/commit/7774c495b7ab4d014a16dc036b284a5b723dc19b

Issue #187, opened 2023-02-28
Allow parse_regex and parse_regex_all to receive as a parameter the result of to_regex like in match function


Genry1:

fpytloun:
I have just the same use-case. Let me share the snippet:

```
        # Support vector annotations to perform parsing
        if is_nullish(parser) {
          parser = get!(.kubernetes.annotations, [join!(["vector.dev/parser", to_string!(.kubernetes.container_name)], separator: "-")])
          if is_nullish(parser) {
            parser = get!(.kubernetes.annotations, ["vector.dev/parser"])
          }
        }

        ## Generic parsers
        if parser == "json" {
          parsed, err = parse_json(.message)
          if err != null {
            log("Unable to parse event with JSON parser:" + err + ". Affected event left unparsed: " + encode_json(.), level: "warn", rate_limit_secs: 10)
          } else {
            . = merge!(to:., from:parsed)
          }
        } else if parser == "regex" {
          reg_ex = get!(.kubernetes.annotations, [join!(["vector.dev/parser", to_string!(.kubernetes.container_name), "regex"], separator: "-")])
          if is_nullish(reg_ex) {
            reg_ex = get!(.kubernetes.annotations, ["vector.dev/parser-regex"])
          }

          reg_ex, err = to_regex(reg_ex)
          if err != null {
            log("Unable to compile regex:" + err + ". Affected event left unparsed: " + encode_json(.), level: "error", rate_limit_secs: 10)
          } else {
            parsed_all, err = parse_regex_all(.message, reg_ex)
            parsed = {}

            for_each(parsed_all) -> |_index, value| {
              parsed = merge(parsed, compact!(value))
            }

            if err != null || parsed == null {
              log("Unable to parse event with regex parser:" + err + ". Affected event left unparsed: " + encode_json(.), level: "warn", rate_limit_secs: 10)
            } else {
              . = merge(to:., from:parsed)
            }
          }
        }
```

PR #302, opened 2023-07-06
feat(regex functions): Allow regex functions to accept variable params

### LINKED ISSUE #114: Allow literal regexes from variables as arguments to VRL functions
A user ran into an issue trying to assign a regex to a variable in VRL and then use that variable in `parse_regex`. This fails because we currently require regex literals but this could be relaxed to allow a variable that contains a regex.

That is, this program should work:

```
my_regex = r'(?<Pfoo>e)'
parse_regex("test", my_regex)
```

KH-Moogsoft:
Intended to resolve: https://github.com/vectordotdev/vrl/issues/114

https://github.com/vectordotdev/vrl/issues/187 also seems to be related.

### DIFF
--- a/src/compiler/function.rs
+++ b/src/compiler/function.rs
@@ -320,6 +320,16 @@ impl ArgumentList {
                 Expr::Literal(crate::compiler::expression::Literal::Regex(regex)) => {
                     Ok((*regex).clone())
                 }
+                Expr::Variable(variable) if variable.value().is_some() => {
+                    match variable.value().unwrap() {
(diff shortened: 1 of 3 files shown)


fuchsnj:
Thank you for your contribution!

While your changes will work here, it is duplicating logic that should live in `Expression::resolve_constant`. The changes I wanted to suggest here are more involved than I wanted to ask an external contributor to do, so I went ahead and made the changes myself [here](https://github.com/vectordotdev/vrl/pull/304). 

I did copy over your examples, so your work here wasn't wasted :)

Issue #151, opened 2023-03-28
Expand constant value support


fuchsnj:
This was started here: https://github.com/vectordotdev/vector/pull/12397

PR #304, opened 2023-07-07
enhancement: Expand usage of `resolve_constant` and add support for variables

fuchsnj:
related: https://github.com/vectordotdev/vrl/issues/151
closes: https://github.com/vectordotdev/vrl/issues/114

`Expr::resolve_constant` tries to resolve an expression with only compile-time information. This implementation has been improved to also resolve variables if the value of the variable is known at compile-time.

This is something that is useful for `ArgumentList` functions, such as `ArgumentList::required_regex`, so for example a variable with a literal regex can be used where a literal regex is required.

All of the `ArgumentList::*` functions were updated to use `Expr::resolve_constant` where approriate, so they can also resolve literal variables (and any other improvements that may come with the `resolve_const` function in the future).

There was also an internal refactor for the `resolve_constant` function. The `TypeState` is the "source of truth" for compile-time information, including variables with known values. The `Variable` expression also included this. This was removed and now the `TypeState` is always used.

### DIFF
--- a/src/compiler/expression/query.rs
+++ b/src/compiler/expression/query.rs
@@ -127,7 +127,9 @@ impl Expression for Query {
 
     fn resolve_constant(&self, state: &TypeState) -> Option<Value> {
         match self.target {
-            Target::Internal(ref variable) => variable.resolve_constant(state),
+            Target::Internal(ref variable) => variable
+                .resolve_constant(state)
+                .and_then(|v| v.get(self.path()).cloned()),
             _ => None,
         }
     }
--- a/src/compiler/expression/query.rs
+++ b/src/compiler/expression/query.rs
@@ -20,22 +20,27 @@ impl Query {
     // TODO:
     // - error when trying to index into object
     // - error when trying to path into array
+    #[must_use]
     pub fn new(target: Target, path: OwnedValuePath) -> Self {
         Query { target, path }
     }
 
+    #[must_use]
     pub fn path(&self) -> &OwnedValuePath {
         &self.path
     }
 
+    #[must_use]
     pub fn target(&self) -> &Target {
         &self.target
     }
 
+    #[must_use]
     pub fn is_external(&self) -> bool {
         matches!(self.target, Target::External(_))
     }
 
+    #[must_use]
     pub fn external_path(&self) -> Option<OwnedTargetPath> {
         match self.target {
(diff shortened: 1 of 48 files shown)


src/compiler/expression/query.rs
```diff
@@ -20,22 +20,27 @@ impl Query {
     // TODO:
     // - error when trying to index into object
     // - error when trying to path into array
+    #[must_use]
```
pront, line 23: Perhaps it's better to apply `#[must_use]` to the `struct Query`.

src/compiler/expression/query.rs
```diff
@@ -117,11 +125,9 @@ impl Expression for Query {
         Ok(value.get(&self.path).cloned().unwrap_or(Value::Null))
     }
 
-    fn resolve_constant(&self) -> Option<Value> {
+    fn resolve_constant(&self, state: &TypeState) -> Option<Value> {
         match self.target {
-            Target::Internal(ref variable) => {
-                variable.value().and_then(|v| v.get(self.path())).cloned()
-            }
+            Target::Internal(ref variable) => variable.resolve_constant(state),
```
pront, line 130: nice

src/compiler/expression/variable.rs
```diff
@@ -13,32 +13,26 @@ use crate::compiler::{
...
 
             return Err(Error::undefined(ident, span, idents));
-        };
+        }
 
-        Ok(Self { ident, value })
+        Ok(Self { ident })
     }
 
+    #[must_use]
```
pront, line 32: Same nit for `#[must_use]`

pront (approved):

src/compiler/expression/query.rs
fuchsnj: Does that mean the same thing? These were from `clippy` suggestions (I'm not sure why it wanted it on these functions all of a sudden)

fuchsnj: I tried it, unfortunately Clippy starts complaining if I only put it on the struct itself :disappointed:

spencergilbert (approved):

src/compiler/expression/query.rs
pront: Ah I see, thanks for trying it out. This might happen because [std::hint::must_use](https://doc.rust-lang.org/std/hint/fn.must_use.html) is still experimental.

PR #332, opened 2023-07-17
prepare for 0.6.0 release

pront:


jszwedko (commented):
There probably isn't any harm in bumping this, but we could wait until we are closer to the Vector 0.32.0 release and bump it then rather than needing to tag 0.7.0 then. What do you think?

pront:
> There probably isn't any harm in bumping this, but we could wait until we are closer to the Vector 0.32.0 release and bump it then rather than needing to tag 0.7.0 then. What do you think?

Sounds good. I will reopen this when we are closer to Vector 0.32.0.

PR #362, opened 2023-08-02
Fix always rebuilding because of build.rs

Zettroke:
Hi, noticed that `vrl` crate was always rebuilding.

Found out the `cargo:rerun-if-changed` had wrong path. Fixed it.


mikhailzagurskiy (approved):

pront (approved):
Thanks @Zettroke!

PR #370, opened 2023-08-04
enhancement: Add `pretty` parameter for `encode_json`

### LINKED ISSUE #358: Add pretty print support for encode_json
Introducing a `pretty` parameter to the `encode_json` function would be beneficial as it enables pretty print support for JSON output. 
When the `pretty` parameter is set to `true`, the function will utilize `serde_json::to_string_pretty` to generate indented and human-readable JSON. 
Conversely, when `pretty` is set to `false` (or not provided), the function will revert to the original behavior of using `serde_json::to_string` for compact JSON output.

Hexta:
Add `pretty` parameter for `encode_json` function to produce pretty-printed JSON string.

closes #358

### DIFF
--- a/src/stdlib/encode_json.rs
+++ b/src/stdlib/encode_json.rs
@@ -1,8 +1,15 @@
 use crate::compiler::prelude::*;
 
-fn encode_json(value: Value) -> Resolved {
+fn encode_json(value: Value, pretty: bool) -> Resolved {
     // With `vrl::Value` it should not be possible to get `Err`.
-    match serde_json::to_string(&value) {
+
+    let result = if pretty {
+        serde_json::to_string_pretty(&value)
+    } else {
+        serde_json::to_string(&value)
+    };
+
+    match result {
         Ok(value) => Ok(value.into()),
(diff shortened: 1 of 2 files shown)


src/stdlib/encode_json.rs
```diff
@@ -1,8 +1,15 @@
 use crate::compiler::prelude::*;
 
-fn encode_json(value: Value) -> Resolved {
+fn encode_json(value: Value, pretty: bool) -> Resolved {
```
pront, line 3: It is a good practice to avoid `bool` parameters. There are many ways to refactor: 

- an options `struct`
- `enum` instead of `bool`
- extract implementation details and split `encode_json` into two functions

pront (approved):
@Hexta: Thank you for submitting this PR!

This LGTM except for the one comment I left.

pront (changes requested):
One more thing, please a new entry in https://github.com/vectordotdev/vrl/blob/main/CHANGELOG.md under `##unreleased`.

pront (approved):

Issue #241, opened 2023-05-13
`undefined` types not handled properly in array literals


fuchsnj:
If you create an array where one of the elements is `undefined`, the actual result is an array with a `null` element, but the type definition thinks that element won't exist (an empty array).

```coffee
$ . = {}
{  }

$ [.x]
[null]

$ type_def([.x])
{ "array": {  } }
```

PR #401, opened 2023-08-18
fix: typedef of array/object literals containing undefined values

fuchsnj:
closes https://github.com/vectordotdev/vrl/issues/241 for both array and object literals


pront (approved):
Interesting.

PR #411, opened 2023-08-24
fix: update AWS VPC Flow Log fields to version 5

thhwalker:
REF: https://github.com/vectordotdev/vrl/issues/227

Updated the fields for `parse_aws_vpc_flow_log` to include the latest version 5 fields. Reference to the new fields can be found here: https://docs.aws.amazon.com/vpc/latest/userguide/flow-logs.html#flow-logs-fields

Added tests from the version 5 example logs: https://docs.aws.amazon.com/vpc/latest/userguide/flow-logs-records-examples.html#flow-log-example-traffic-path

Ran checks and confirmed passed.


pront (approved):
These changes look reasonable.

Issue #263, opened 2023-06-02
bug: parse_aws_vpc_flow_log types account-id as int64 instead of string


timcosta:
Hi all! I've been parsing some VPC Flow Logs with VRL, and noticed that a chunk of them are being sidelined due to errors. All of the sidelined logs have `unknown` as the value for the `account-id` column, which is [listed as possible in the AWS docs](https://docs.aws.amazon.com/vpc/latest/userguide/flow-logs.html#flow-logs-default).

VRL [appears to type account-id as an int64](https://github.com/vectordotdev/vrl/blob/2bbe6728fc0a9d0c20b6ee8634346bdbf28f78d8/src/stdlib/parse_aws_vpc_flow_log.rs#L194), which makes it fail to parse any VPC Flow logs where the account id is unknown. I believe parsing as int64 will also drop any leading `0` characters, which would make the output incorrect for any account id that has one or more leading zeroes.

Sample log with ENI and IPs obfuscated:
```
2 unknown eni-070b1422222222222 192.168.1.1 10.0.0.2 55828 60002 6 1 44 1685727399 1685727446 ACCEPT OK
```

Let me know if you need any additional information!

thhwalker:
Just thinking out loud here, but would it be feasible to mark the account ID as a string in this instance?
AWS have specified the key as a string type, and at face value I don't see any specific need for this to be an integer for the sake of manipulating the value.

That being said, I'm cognisant of the fact that there may be users that are handling these values specifically as integers. I'm not sure what they would be doing with this information, but I feel it's something to be mindful of.

I could just be overthinking this one.

I'm more than happy to handle a PR for this if changing `Kind::integer` to `Kind::bytes` is the agreed solution.

jszwedko:
Yeah, agreed, this seems like it should be a string rather than an `int64`.

PR #415, opened 2023-08-25
fix: vpc flow log field account-id as string

thhwalker:
REF: https://github.com/vectordotdev/vrl/issues/263

Changed field `account_id` from an Integer type to a string.
As stated in [VPC Flow Log documentation](https://docs.aws.amazon.com/vpc/latest/userguide/flow-logs.html#flow-logs-fields), account-id is labelled as a string type. It is also stated that this field may have a value of `unknown`. Some account ID's may have leading zeros, which would be dropped during conversion to integer.


pront (approved):
LGTM!

PR #416, opened 2023-08-25
chore(deps): Update to Rust 1.72.0

pront:


src/stdlib/find.rs
```diff
@@ -1,5 +1,6 @@
 use crate::compiler::prelude::*;
 
+#[allow(clippy::cast_possible_wrap)]
```
pront, line 3: Note to reviewer: 

I had a few ideas here on how to fix but I didn't want to change behavior in a deps update kind of PR. Let me know if you have other ideas. Btw, do we document expected wrapping behavior for arithmetic expressions?

jszwedko: 👍 to following up on this lint outside of this PR.

It seems like this overflow could only happen when the value in the returned `usize` is outside of the bounds of an `i64`. We probably do need to define a general rule of how to handle this sort of overflow in VRL. I'm open to ideas.

jszwedko (approved):
Great!

src/stdlib/find.rs
fuchsnj: I'm fine with either adding a comment here saying it can't actually happen, or maybe using a fallible conversion and using `unreachable`. A string can't actually get long enough due to memory constraints, but https://github.com/vectordotdev/vrl/issues/242 will also add a most realistic upper bound.

fuchsnj (approved):

src/stdlib/find.rs
pront: I will follow-up with a small PR on that. It would be good to update the public docs for overflows.

PR #442, opened 2023-09-08
Loosen Nginx error log regex matching

Leo1003:
Since the fields can be empty in some circustances, it is better to use '*' instead of '+' to match the values.

Below is an example log from our Nginx container which can fail the `parse_nginx_log()` finction:
```
2023/09/08 13:50:28 [warn] 3#3: *531 an upstream response is buffered to a temporary file /var/lib/nginx/tmp/fastcgi/6/03/0000000036 while reading upstream, client: 140.113.168.124, server: , request: "GET / HTTP/1.1", upstream: "fastcgi://127.0.0.1:9000", host: "duty.cs.nctu.edu.tw", referrer: "https://oauth.cs.nctu.edu.tw/"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## unreleased
 
+- `parse_nginx_log` doesn't fail if the values of key-value pairs in error logs is missing (https://github.com/vectordotdev/vrl/pull/442)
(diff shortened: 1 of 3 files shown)


pront:
Hi @Leo1003, thanks for submitting this change! Can you please add a test and an entry in the CHANGELOG.md?

Leo1003:
I have added a test and an entry in the CHANGELOG

pront (approved):
Thank you!

Issue #417, opened 2023-08-27
contains all the given substrings in string (contains_all)


saeed-mcu:
I want to determines whether the `string` matches **all** the given `substrings` or not. (all of them MUST contains)
Is it possible to all substrings are checked in a single pass over the target string to achieve good performance advantage over multiple calls to single `contains` function.

if **`contains_all`** function added in String functions of VRL , it can be very useful and applicable for log parsing.

pront:
Hi @saeed-mcu, is the following example what you were thinking?

```
contains_all("The Needle In The Haystack", "Needle", "Haystack") # returns true
```

If yes, that seems reasonable.

saeed-mcu:
Hi @pront  , is the following example return false ?
```
contains_all("The Needle In The Haystack", "Needle", "not_exist") 
```
I expected , all substring exist in the string not one of them.
so I expected above example return false

Issue #458, opened 2023-09-18
Explore support for a native code transform 


pront:
The following is not a formal proposal/design but just an idea that could be explored in the future.
 
Today Vector supports the [Lua transform](https://vector.dev/docs/reference/configuration/transforms/lua/). This issue is a suggestion to explore something similar either as part of VRL. I am adding this issue as placeholder for potential future work.

Let's use string manipulation as a motivating example.

Imagine a user wants to determine whether a string contains all the following substrings ["abc", "xyz"]. This can be done today by using multiple VRL functions and combining their results. See the following VRL snippet:

```
sentence = "A needle in the haystack"
contains_needle = contains(sentence, "needle")
contains_haystack = contains(sentence, "haystack")
contains_both = contains_needle && contains_haystack

if contains_both {
    # do A
} else {
    # do B
}
```

If VRL had a transform that supported Rust code it would look like:
```
sentence = "A needle in the haystack"
contains_both = contains_all_native(sentence, "needle", "haystack")

if contains_both {
    # do A
} else {
    # do B
}
```

This allows for tremendous flexibility and reduces engineering efforts since we don't need to introduce a new VRL function for each string manipulation that needs optimization.

Note that the user would also need to implement `contains_all_native` and that would need to be compiled first and then invoked. There are several unanswered design questions here.

Issue #261, opened 2023-05-26
`parse_groks` aliases are not type checked


fuchsnj:
The following code will panic. The 3rd argument is an object containing grok aliases. Each value should be a string, but the `parse_groks` function fails to check this.

```coffee
parse_groks!("", [], {"foo": 3})
```

VRL version: `main` (ebb1f2b739e776)

PR #468, opened 2023-09-20
feat: Add 'contains_all' VRL function

pront:
closes: #417 https://github.com/vectordotdev/vrl/issues/417

# Summary
* Adds new function `contains_all` function that accepts multiple substrings and checks against each of them.
* Removed some code duplication in string manipulation functions.


src/stdlib/string_utils.rs
```diff
@@ -0,0 +1,9 @@
+use crate::prelude::{Value, ValueError, VrlValueConvert};
+
+pub(crate) fn convert_to_string(value: Value, case_sensitive: bool) -> Result<String, ValueError> {
```
fuchsnj, line 3: `case_sensitive` doesn't really make sense here. It's not clear if it's converting to lowercase or uppercase. Maybe just call it `to_lowercase: bool`?

src/stdlib/contains_all.rs
```diff
@@ -0,0 +1,152 @@
+use crate::compiler::prelude::*;
+use crate::stdlib::string_utils::convert_to_string;
+
+fn contains_all(value: Value, substrings: Value, case_sensitive: bool) -> Resolved {
```
fuchsnj, line 4: Is it worth making this function a bit more generic? I could see someone asking for a `contains_any` for the same reasons as `contains_all`.

fuchsnj: We could also look into using https://crates.io/crates/aho-corasick for a significant perf boost when used with long strings. That algorithm is specifically designed for this.

src/stdlib/contains_all.rs
```diff
@@ -0,0 +1,152 @@
...
+
+    fn compile(
+        &self,
+        _state: &state::TypeState,
+        _ctx: &mut FunctionCompileContext,
+        arguments: ArgumentList,
+    ) -> Compiled {
+        let value = arguments.required("value");
+        let substrings = arguments.required("substrings");
+        let case_sensitive = arguments.optional("case_sensitive").unwrap_or(expr!(true));
```
fuchsnj, line 53: I generally prefer making the expression optional in `ContainsAllFn` over creating a new `true` expression, but it's not a big deal here. In general it's faster to have an optional so you don't have to evaluate an expression when we are using the "default" value.

src/stdlib/contains_all.rs
pront: Yes, good point. This is what inspired: https://github.com/vectordotdev/vrl/issues/458. We could make this more generic by adding OR/AND semantics.

For the perf work, I generally like to do this as a secondary step. I will look into it.

pront: Discussed offline: 

- keeping this as is
- we can add `contains_any` in the future if there's demand
  - `aho-corasick` seems a good candidate there

src/stdlib/contains_all.rs
```diff
@@ -0,0 +1,161 @@
...
+    fn parameters(&self) -> &'static [Parameter] {
+        &[
+            Parameter {
+                keyword: "value",
+                kind: kind::BYTES,
+                required: true,
+            },
+            Parameter {
+                keyword: "substrings",
+                kind: kind::ARRAY,
```
fuchsnj, line 39: I almost missed this one. The `substrings` input is only guaranteed to be an array when called, not necessarily an array _of strings_. As written, this function should be fallible since the `convert_to_string` could fail. I'd suggest adding a check in `compile` function.

fuchsnj: See this related issue: https://github.com/vectordotdev/vrl/issues/261
I keep seeing this come up. We should probably find a better way to prevent this since it's easy to miss. Fuzz testing can catch this, but maybe the `kind::ARRAY` should be changed to allow specifying the element type, or something like that.

pront: Updated the code, I like the idea of specifying the ARRAY type. I will make an issue.

fuchsnj (approved):

PR #470, opened 2023-09-21
chore: standard cargo commands work without requiring the 'test' feature

pront:
We can now run `cargo test` or `cargo clippy` without `--features "tests"`.


jszwedko (commented):
I didn't review the changes, but I just wanted to say I'm a big 👍 to changes in this vein since they reduce cognitive load for people.

bruceg (approved):

PR #484, opened 2023-09-30
Add parse_float() VRL function

### LINKED ISSUE #266: Add function to parse float
Currently, there is no function can simply parse float string.

Maybe, something like `parse_int()` should be added.

Leo1003:
New VRL function  `parse_float()` to parse float string. Using `f64::from_str()` in the standard library as underlying implementation.

Close #266

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,6 @@
 
 ## unreleased
 
-#### Features
-- 'from_unix_timestamp' now accepts a new unit: Microseconds.
-
 ## `0.7.0` (2023-09-25)
 
 #### Bug Fixes
@@ -21,8 +18,6 @@
 - removed deprecated `to_timestamp` function (https://github.com/vectordotdev/vrl/pull/452)
 - changed `truncate` arguments, it now accepts a suffix string instead of a boolean (https://github.com/vectordotdev/vrl/pull/454) 
 
-- added `contains_all` function (https://github.com/vectordotdev/vrl/pull/468)
-- 
 ## `0.6.0` (2023-08-02)
 
 #### Bug Fixes
--- a/src/stdlib/parse_float.rs
+++ b/src/stdlib/parse_float.rs
@@ -1,0 +1,142 @@
+use crate::compiler::prelude::*;
+use std::str::FromStr;
+
+fn parse_float(value: Value) -> Resolved {
+    let string = value.try_bytes_utf8_lossy()?;
+    let float = f64::from_str(&string)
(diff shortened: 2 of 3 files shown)


src/stdlib/parse_float.rs
```diff
@@ -0,0 +1,142 @@
...
+            want: Ok(f64::NEG_INFINITY),
+            tdef: TypeDef::float().fallible(),
+        }
+
+        nan {
+            args: func_args![value: "Nan"],
+            want: Err("NaN number not supported".into()),
+            tdef: TypeDef::float().fallible(),
+        }
+    ];
```
pront, line 141: Let's add two cases for [MAX](https://doc.rust-lang.org/std/primitive.f64.html#associatedconstant.MAX) and [MIN](https://doc.rust-lang.org/std/primitive.f64.html#associatedconstant.MIN). Since these special values tend to be a common source of bugs.

pront (changes requested):
Hi @Leo1003, thank you for this PR! 
It looks good overall. Left a couple a comments.

src/stdlib/parse_float.rs
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    fn parameters(&self) -> &'static [Parameter] {
+        &[
+            Parameter {
+                keyword: "value",
+                kind: kind::BYTES,
+                required: true,
+            },
+        ]
```
pront, line 29: To make this function a bit more generic, let's add a base parameter like we do in [parse_int](https://github.com/vectordotdev/vrl/blob/main/src/stdlib/parse_int.rs#L49).

Leo1003: Since the Rust standard library doesn't support parsing float with other radix. To support this, we have to implement the whole parsing logic by ourselves. Also, I think it is very rare to see float numbers in other radix. In my opinion, it is not worthy to support the base parameter.

pront: This makes sense, I agree. We can go ahead with this PR as is. Note, there are some failing checks.

pront (approved):
Thank you for this contribution @Leo1003!

CHANGELOG.md
```diff
@@ -2,10 +2,11 @@
 
 ## unreleased
 
-- `parse_nginx_log` no longer fails if `upstream_response_length`, `upstream_response_time`, `upstream_status` are missing (https://github.com/vectordotdev/vrl/pull/498)
-
 #### Features
+- added `contains_all` function (https://github.com/vectordotdev/vrl/pull/468)
```
pront, line 6: Fixed base rebase.

PR #489, opened 2023-10-03
Enhancement: Conversion for TimeZone to FixedOffset

kates:
To aid easier conversion of VRL's TimeZone to FixedOffset as requested in https://github.com/vectordotdev/vector/pull/18506

### DIFF
--- a/src/compiler/datetime.rs
+++ b/src/compiler/datetime.rs
@@ -1,7 +1,7 @@
 use std::fmt::Debug;
 
 use chrono::format::{parse, Parsed, StrftimeItems};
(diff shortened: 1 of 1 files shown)


src/compiler/datetime.rs
```diff
@@ -84,3 +84,12 @@ impl TryFrom<String> for TimeZone {
         }
     }
 }
+
+impl From<TimeZone> for FixedOffset {
```
pront, line 88: also see my [comment here](https://github.com/vectordotdev/vector/pull/18506/files#r1344284369)

kates: Thanks @pront, I've implemented the conversion to remove duplication in the Vector repo.

PR #492, opened 2023-10-03
feat(stdlib): support microseconds in from_unix_timestamp

### LINKED ISSUE #491: Support microseconds in from_unix_timestamp
Hi,

As of version 0.4.31, chrono has a function that takes a UNIX timestamps in microseconds and returns a DateTime struct: [Utc.timestamp_micros](https://docs.rs/chrono/latest/chrono/offset/trait.TimeZone.html#method.timestamp_micros).
I want to add support for using UNIX microsecond timestamps in `from_unix_timestamp.rs`. I forked the repository and implemented my request [here](https://github.com/vectordotdev/vrl/compare/main...final-israel:vrl:support_microseconds_in_from_unix_timestamp).

With this, would vector users be able to use the new `Unit::Microseconds` in their vrl remaps?
Is my implementation acceptable?

emikitas:
resolves #491

### DIFF
--- a/src/stdlib/from_unix_timestamp.rs
+++ b/src/stdlib/from_unix_timestamp.rs
@@ -7,20 +7,18 @@ fn from_unix_timestamp(value: Value, unit: Unit) -> Resolved {
 
     let value = match value {
         Integer(v) => match unit {
-            Unit::Seconds => {
-                let t = Utc.timestamp_opt(v, 0).single();
-                match t {
-                    Some(time) => time.into(),
-                    None => return Err(format!("unable to coerce {v} into timestamp").into()),
-                }
-            }
-            Unit::Milliseconds => {
-                let t = Utc.timestamp_millis_opt(v).single();
-                match t {
-                    Some(time) => time.into(),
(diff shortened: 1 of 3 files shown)


src/stdlib/from_unix_timestamp.rs
```diff
@@ -198,6 +205,12 @@ mod tests {
             tdef: TypeDef::timestamp().fallible(),
         }
 
+        integer_microseconds {
```
pront, line 208: Let's also add an example (line 74).

pront (approved):
Thank you for this contribution @emikitas. This looks good.

emikitas:
I noticed I botched one of my commit messages, so I decided to squash the commit adding the example into it and amend the commit message.

pront:
> I noticed I botched one of my commit messages, so I decided to squash the commit adding the example into it and amend the commit message.

Hi @emikitas, feel free to push as many commits as you need. In the end will squash all those. However, make sure to follow https://www.conventionalcommits.org/en/v1.0.0/ for the final commit.

PR #498, opened 2023-10-09
enhancement: allow more missing upstream info in parse_nginx_log

ex5:
There's a valid use case when `upstream_response_length`, `upstream_response_time`, `upstream_status` are sometimes unavailable but the rest of the log message still matches the `upstreaminfo` format (e.g. static files served by the same virtual host). Here's an example of such a message:

```
1.1.1.1 - - [09/Oct/2023:10:49:43 +0200] "GET /media/cache/f4/b4/f4b494deadbeefc8.jpg HTTP/1.1" 200 32235 "https://example.com/page/" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/117.0.0.0 Safari/537.36" 951 0.000 [upstream-production-service-1234] [upstream-production-service-1234] - - - - 7d1f3deadbeefcbb3bc
```

Currently, this kind of message generates a parsing error and some logging data is lost. It would be useful if instead the rest of the available data was still parsed correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## unreleased
 
-- `parse_nginx_log` doesn't fail if `upstream_response_length`, `upstream_response_time`, `upstream_status` are missing (https://github.com/vectordotdev/vrl/pull/498)
+- `parse_nginx_log` no longer fails if `upstream_response_length`, `upstream_response_time`, `upstream_status` are missing (https://github.com/vectordotdev/vrl/pull/498)
 
 #### Features
 - 'from_unix_timestamp' now accepts a new unit: Microseconds.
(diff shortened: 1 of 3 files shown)


pront (changes requested):
Hi @ex5, this change seems reasonable to me, thanks! 

Two things we still need:
* Add a changelog.md entry for this change
* Update the Vector docs [here](https://github.com/vectordotdev/vector/blob/master/website/cue/reference/remap/functions/parse_nginx_log.cue) with a new example with arguments like the  `ingress_nginx_upstreaminfo_valid_missing_upstream`. 
  * `cue fmt path/to/file` and `make check-component-docs` are the commands you will need.

pront (approved):
Looks good, thank you.

PR #523, opened 2023-10-24
feat(diagnostics): improve fallibility error notes

pront:
part of https://github.com/vectordotdev/vrl/issues/23

### DIFF
--- a/lib/tests/tests/diagnostics/unhandled_parse_regex_all_type.vrl
+++ b/lib/tests/tests/diagnostics/unhandled_parse_regex_all_type.vrl
@@ -9,10 +9,12 @@
 #   │ │    │
 #   │ │    this expression is fallible
 #   │ │    update the expression to be infallible
+#   │ │    note that argument types can affect fallibility
 #   │ or change this to an infallible assignment:
 #   │ .a, err = sha3(.result[0].an)
 #   │
 #   = see documentation about error handling at https://errors.vrl.dev/#handling
+#   = see functions characteristics documentation at https://vrl.dev/expressions/#function-call-characteristics
 #   = learn more about error code 103 at https://errors.vrl.dev/103
 #   = see language documentation at https://vrl.dev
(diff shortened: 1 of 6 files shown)


lib/tests/tests/diagnostics/unhandled_parse_regex_all_type.vrl
```diff
@@ -9,10 +9,12 @@
 #   │ │    │
 #   │ │    this expression is fallible
 #   │ │    update the expression to be infallible
+#   │ │    note if an argument type is invalid it can render a function fallible
```
jszwedko, line 12: I think it would be great if we could go even further and say, for cases where the function is fallible due to the argument type not being known to be valid, that:

```
this expression is fallible because argument `.result[0].an` is of type `any` and this function expects a `string`. It will fail at runtime if `.result[0].an` is not a `string`
```

The currently added message is definitely an improvement, but I think users will remain largely confused by this behavior of VRL.

pront: @jszwedko: Agreed, good point. 

However, I don't have time to do it before this upcoming release. Two options:

1. Get this improvement in and follow up with another PR (to be included in the next release)
2. Leave this PR open

What do you prefer?

jszwedko: 👍 I'm fine with merging this since it is an improvement and then following up. I just want to make sure we don't close out https://github.com/vectordotdev/vrl/issues/23 until we do that.

pront: Sounds good. Added a reminder [here](https://datadoghq.atlassian.net/browse/OPW-89?focusedCommentId=1390837). 

Merging this PR won't close the ticket, so if you give this a ✅ I will include it in the VRL release.

jszwedko (approved):

PR #526, opened 2023-10-25
feat: fallibility is now an enum

pront:
closes: https://github.com/vectordotdev/vrl/issues/32

### DIFF
--- a/src/compiler/type_def.rs
+++ b/src/compiler/type_def.rs
@@ -40,10 +40,17 @@ pub enum Fallibility {
 
 impl Fallibility {
     #[must_use]
-    pub fn merge(&self, other: &Self) -> Self {
+    /// Merges two [`Fallibility`] values using the following rules:
+    ///
+    /// - Merging with [`Fallibility::AlwaysFails`] always results in [`Fallibility::AlwaysFails`].
+    /// - Merging [`Fallibility::MightFail`] with any variant results in [`Fallibility::MightFail`].
+    /// - Merging two [`Fallibility::CannotFail`] values results in [`Fallibility::CannotFail`].
+    ///
+    /// This is useful for combining the fallibility of sub-expressions.
+    pub fn merge(left: &Self, right: &Self) -> Self {
         use Fallibility::{AlwaysFails, CannotFail, MightFail};
 
-        match (self, other) {
+        match (left, right) {
             (AlwaysFails, _) | (_, AlwaysFails) => AlwaysFails,
             (MightFail, _) | (_, MightFail) => MightFail,
             _ => CannotFail,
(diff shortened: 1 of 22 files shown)


src/compiler/type_def.rs
```diff
@@ -31,6 +31,26 @@ use crate::value::{
...
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum Fallibility {
+    CannotFail,
+    MightFail,
+    AlwaysFails,
+}
+
+impl Fallibility {
+    #[must_use]
+    pub fn merge(&self, other: &Self) -> Self {
```
fuchsnj, line 43: This one needs some documentation, or potentially a more descriptive name. Without reading all the code that uses this, it's not clear what "merging" fallibility means. It might mean you have 2 expressions executed sequentially, and you are combining the fallibility of both sub-expressions (which seems to match the logic here), but there's other ways this could be merged, such as 2 parallel branches (think of an if statement, where the fallibility of each branch is different, and you want to merge, the logic will be different in that case)

src/stdlib/slice.rs
```diff
@@ -109,6 +110,58 @@ struct SliceFn {
...
+/// After going through the deprecation process, we can use this to produce
+/// a more accurate type definition and make more `slice` uses infallible.
+fn determine_fallibility(
+    value: &dyn Expression,
+    start: &dyn Expression,
+    end: Option<&dyn Expression>,
+    state: &state::TypeState,
+) -> Fallibility {
+    let array = match value.resolve_constant(state) {
+        None => return Fallibility::AlwaysFails,
```
fuchsnj, line 125: This is wrong. If `resolve_constant` returns `None`, that just means the value is unknown at compile time, but it's certainly possible for it to be successful still. It seems like there are similar issues elsewhere too.

fuchsnj: In general I feel like we can add several helper functions to simplify this logic. The `determine_fallibility` function seems quite verbose right now. I'll try suggesting some if you can't come up with any (after the issue above is fixed)

src/compiler/type_def.rs
```diff
@@ -39,7 +59,7 @@ pub struct TypeDef {
     ///
     /// Some expressions are infallible (e.g. the [`Literal`][crate::expression::Literal] expression, or any
     /// custom function designed to be infallible).
-    fallible: bool,
+    fallibility: Fallibility,
```
fuchsnj, line 62: We should document the requirement for accuracy of the `Fallibility` somewhere. Maybe on the `Expression::type_def` function, or maybe here? Something along the lines of, if the function _can_ fail, the fallibility must _not_ be `CannotFail`, and if the function _can_ succeed, the fallibility must not be `AlwaysFails`. The rest is best effort. (So always returning `MightFail` is "correct", but potentially undesirable.

src/stdlib/slice.rs
pront: Thanks for the suggestions. It all makes sense. I will do this another PR in order to unblock this one.

pront:
> Documenting some good point from our offline discussion for potential followups.

Consider the follow example:
```
if (some condition) { AlwaysFails } else { CannotFail }
```
^ We can assign a fallibility to each expression and bubble it up (a single `AlwaysFail` is enough for the whole program to fail).

Nathan (fuchsnj) made a great point that an unnecessary `!` should not be treated as a hard error. If we relax that, then we take advantage of new fallibility checks without breaking existing VRL programs.

fuchsnj (approved):

PR #530, opened 2023-10-26
chore: Add wrapper type for key strings

bruceg:
This change adds a newtype struct for the object key type called `KeyString` and a convenience type alias for the object mapping type called `ObjectMap`. This lays the groundwork for future changes to the key data model.

### DIFF
--- a/src/core/encode_key_value.rs
+++ b/src/core/encode_key_value.rs
@@ -154,7 +154,7 @@ impl fmt::Display for Data {
 }
 
 struct KeyValueSerializer<'a> {
-    key: String,
+    key: KeyString,
     separator: char,
     output: &'a mut BTreeMap<KeyString, Data>,
 }
@@ -162,7 +162,7 @@ struct KeyValueSerializer<'a> {
 impl<'a> KeyValueSerializer<'a> {
     fn new(key: KeyString, separator: char, output: &'a mut BTreeMap<KeyString, Data>) -> Self {
         Self {
-            key: key.into(),
+            key,
             separator,
             output,
         }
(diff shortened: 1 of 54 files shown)


pront:
I will review this today and hopefully we can get it merged before the next VRL release.

src/core/encode_key_value.rs
```diff
@@ -154,13 +156,13 @@ impl fmt::Display for Data {
...
-    output: &'a mut BTreeMap<String, Data>,
+    output: &'a mut BTreeMap<KeyString, Data>,
 }
 
 impl<'a> KeyValueSerializer<'a> {
-    fn new(key: String, separator: char, output: &'a mut BTreeMap<String, Data>) -> Self {
+    fn new(key: KeyString, separator: char, output: &'a mut BTreeMap<KeyString, Data>) -> Self {
         Self {
-            key,
+            key: key.into(),
```
pront, line 165: Is it possible to implicitly convert here?

pront (approved):
LGTM. My only non blocking question is if we can go further and avoid using `.into()` everywhere.

src/core/encode_key_value.rs
bruceg: Oh, good catch. The field in the struct should have been a `KeyString` as seen below in `fn process` this results in a double conversion. Fixed in f0c79084b

src/datadog/grok/parse_grok_rules.rs
```diff
@@ -270,7 +270,7 @@ fn resolve_grok_pattern(
     }
 
     let match_name = &pattern.match_fn.name;
-    match context.aliases.get(match_name).cloned() {
+    match context.aliases.get(match_name.as_str()).cloned() {
```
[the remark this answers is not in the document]
bruceg: We [already do](https://github.com/vectordotdev/vrl/pull/530/files/c2f0b4c6afb25df26928ac9dbdb58430e89f1f50#diff-a4c361e029b3a0812d7b832ab810fcd377851e901ffda8955d706e09e03d31fcR54), actually. I don't see how it helps here. `match_name` is a `String` and we need to borrow it as a `KeyString`. Rust doesn't know how to do that directly, so this needs a `.as_str()` to provide the intermediate between the two.

bruceg:
> My only non blocking question is if we can go further and avoid using `.into()` everywhere.

Some of those are a feature, actually, to make explicit we are converting from some kind of string into a `KeyString`. Right now, that is just adding the wrapper type, but eventually we will be playing with the internals, making those `into`s necessary.

pront (approved):
Now that we published a new VRL release, we can get this merged.

PR #539, opened 2023-10-31
fix: build dep 'lalrpop' specified a yanked version

pront:


Cargo.lock
```diff
@@ -2960,7 +2960,7 @@ checksum = "49874b5167b65d7193b8aba1567f5c7d93d001cafc34600cee003eda787e483f"
 
 [[package]]
 name = "vrl"
-version = "0.7.0"
+version = "0.8.0"
```
jszwedko, line 2963: Should this change have been in https://github.com/vectordotdev/vrl/pull/538?

Cargo.toml
```diff
@@ -155,7 +155,7 @@ regex = { version = "1", default-features = false, features = ["std", "perf", "u
 paste = { version = "1", default-features = false }
 
 [build-dependencies]
-lalrpop = { version = "0.20.1", default-features = false }
+lalrpop = { version = "0.20", default-features = false }
```
jszwedko, line 158: The lock file still seems to be using `0.20.1`. Did you maybe need to run `cargo update -p lalrpop`?

Cargo.lock
pront: Yes, talked to @fuchsnj and we yanked the VRL version 0.8.0.

Cargo.toml
pront: Yes, updated in https://github.com/vectordotdev/vrl/pull/539/commits/e8fc747a9d73c7b181ea1f066b4c8cb37f730cc2

jszwedko (approved):

Issue #453, opened 2023-09-15
Error message loop with variables in closures


lukesteensen:

pront:
This is adjacent to the fallibility diagnostics improvements that we are planning. A potential solution is to add a [fallibility property](https://github.com/vectordotdev/vrl/pull/526) to each expression (not just functions) and then by traversing the AST we can determine which expressions are causing the code block compilation failure.

PR #553, opened 2023-11-13
feat(diagnostics): improve assignment fallibility compilation error

pront:
closes: #23

### DIFF
--- a/lib/tests/tests/diagnostics/unhandled_parse_regex_all_type.vrl
+++ b/lib/tests/tests/diagnostics/unhandled_parse_regex_all_type.vrl
@@ -9,7 +9,7 @@
 #   │ │    │
 #   │ │    this expression is fallible
 #   │ │    update the expression to be infallible
-#   │ │    '.result[0].an' argument type is 'string or undefined' and this function expects parameter 'value' of type 'string'
+#   │ │    '.result[0].an' argument type is 'string or undefined' and this function expected a parameter 'value' of type 'string'
 #   │ or change this to an infallible assignment:
 #   │ .a, err = sha3(.result[0].an)
 #   │
--- a/lib/tests/tests/diagnostics/unhandled_parse_regex_all_type.vrl
+++ b/lib/tests/tests/diagnostics/unhandled_parse_regex_all_type.vrl
@@ -9,7 +9,7 @@
 #   │ │    │
 #   │ │    this expression is fallible
 #   │ │    update the expression to be infallible
-#   │ │    '.result[0].an' argument type is unknown and this function expects parameter 'value' of type 'string'
+#   │ │    '.result[0].an' argument type is 'string or undefined' and this function expects parameter 'value' of type 'string'
 #   │ or change this to an infallible assignment:
 #   │ .a, err = sha3(.result[0].an)
 #   │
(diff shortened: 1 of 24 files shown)


src/compiler/expression_error.rs
```diff
@@ -40,14 +41,14 @@ impl From<ExpressionError> for Diagnostic {
     }
 }
 
-impl DiagnosticMessage for ExpressionError {
+impl DiagnosticMessage for ExpressionError2 {
```
fuchsnj, line 44: This seems like a temporary name, is it supposed to change to something else, or are we keeping it for a bit?

fuchsnj (approved):

src/compiler/expression_error.rs
pront: Addressed here: https://github.com/vectordotdev/vrl/pull/554

lib/tests/tests/diagnostics/unhandled_parse_regex_all_type.vrl
```diff
@@ -9,7 +9,7 @@
 #   │ │    │
 #   │ │    this expression is fallible
 #   │ │    update the expression to be infallible
-#   │ │    note if an argument type is invalid it can render a function fallible
+#   │ │    '.result[0].an' argument type is 'string or undefined' and this function expected a parameter 'value' of type 'string'
 #   │ or change this to an infallible assignment:
 #   │ .a, err = sha3(.result[0].an)
```
jszwedko, line 14:
```
#   │ │    this expression is fallible because at least one argument's type cannot be verified to be valid
#   │ │    '.result[0].an' argument type is 'string or undefined' and this function expected a parameter 'value' of type 'string'
#   │  update the expression to be infallible by adding a `!`: `.a = sha3!(.result[0].an)`
#   │ or change this to an infallible assignment:
#   │ .a, err = sha3(.result[0].an)
```

What you have here is a great step forward, but do you think we can make it even clearer with something like what I put? I still worry that the current error message will be confusing to users. I'm not sure exactly what's possible though.

pront: If I read the above correctly you want to add another line:
```
update the expression to be infallible by adding a `!`: `.a = sha3!(.result[0].an)`
```

which should be doable and I will follow up on this.

Ultimately users should have some basic understanding about [fallibility](https://vector.dev/docs/reference/vrl/#fallibility) and [fail safety](https://vector.dev/docs/reference/vrl/#fail-safety) when using VRL (which practically means adding a `!` or handling the returned `err`).

P.S. Where users might still be confused, is with code blocks which contain fallible expressions [example here](https://github.com/vectordotdev/vrl/issues/453). However, that part of the codebase is not easy to maintain and is need of refactoring.

jszwedko:
```
#   │ │    this expression is fallible because at least one argument's type cannot be verified to be valid
#   │ │    '.result[0].an' argument type is 'string or undefined' and this function expected a parameter 'value' of type 'string'
```

is actually the more important update I would like to see. I think my verbiage makes it more clear exactly what the issue is: that the function is fallible because one of the arguments cannot be verified, and this is the argument (or arguments) in particular, their types, and what the function expects.

The hints on how to fix are then a cherry on top.

pront: Gotcha, this makes sense.

PR #560, opened 2023-11-14
feat(diagnostics): improve fallibility assignment error

pront:
closes: #23


jszwedko (approved):
Awesome! I think this is a great improvement.

Issue #562, opened 2023-11-15
Unable to pass elements of regular expressions list to function parse_regex_all


spacepatcher:
Hi there!
I have a list of regular expressions exported from an external source. I want to use the function `parse_regex_all` with each of them as a parameter. I'm looking for a way to create a code that follows this logic:
```
regex_list = [
r'pattern1',
r'pattern2'
]

for_each(regex_list) -> |_index, _| {
    .http_status = to_string!(.http_status)

    parse_regex_all!(.http_status, regex_list[_index]) # 1. 
    parse_regex_all!(.http_status, get!(regex_list, [_index])) # 2. 
}
```
I expected the function calls in `1.` or `2.` to be executed without any errors, but in fact, errors have occurred accordingly:
```
error[E203]: syntax error
   ┌─ :11:47
   │
11 │     parse_regex_all!(.http_status, regex_list[_index])
   │                                               ^^^^^^
   │                                               │
   │                                               unexpected syntax token: "Identifier"
   │                                               expected one of: "integer literal"
   │
   = see language documentation at https://vrl.dev
   = try your code in the VRL REPL, learn more at https://vrl.dev/examples
```

```
error[E610]: function compilation error: error[E400] unexpected expression type
   ┌─ :13:5
   │
13 │     parse_regex_all!(.http_status, get!(regex_list, [_index]))
   │     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   │     │
   │     unexpected expression for argument "pattern"
   │     expected: regex
   │     received: function call
   │
   = learn more about error code 400 at https://errors.vrl.dev/400
   = see language documentation at https://vrl.dev
   = try your code in the VRL REPL, learn more at https://vrl.dev/examples
```

Would you give me any clues on how to implement the desired logic. Cheers!

bruceg:
When using `for_each`, the second value passed to the closure is the value in the array. So, you don't need to use the index for what you want:
```coffee
.http_status = to_string!(.http_status)
for_each(regex_list) -> |_, pattern| {
    parse_regex_all!(.http_status, pattern)
}
```

spacepatcher:
Yes, I also tried it this way, but also without any result. The compiler is very strict and breaks off with a familiar error:
```
error[E610]: function compilation error: error[E400] unexpected expression type
  ┌─ :8:5
  │
8 │     parse_regex_all!(.http_status, pattern)
[...]

PR #563, opened 2023-11-16
feat(parse_regex_all): allow variables for the 'pattern' parameter

pront:
closes: #562 

Tested with 
```
regex_list = [ r'pattern1', r'pattern2' ]

result = {}
for_each(regex_list) -> |i, pattern| {
    parse_result = parse_regex_all!(to_string!(.http_status), pattern, numeric_groups: true)
    if length(parse_result) > 0 {
        result = set!(result, [i], parse_result)
    }
}
. = result
```

Result
```
[[{ "0": "pattern1" }], [{ "0": "pattern2" }]]
```

### DIFF
--- a/src/stdlib/parse_regex_all.rs
+++ b/src/stdlib/parse_regex_all.rs
@@ -1,5 +1,6 @@
-use crate::compiler::prelude::*;
 use regex::Regex;
+
+use crate::compiler::prelude::*;
 
 use super::util;
 
@@ -30,7 +31,7 @@ impl Function for ParseRegexAll {
             },
             Parameter {
                 keyword: "pattern",
-                kind: kind::ANY,
+                kind: kind::REGEX,
                 required: true,
             },
             Parameter {
@@ -43,12 +44,12 @@ impl Function for ParseRegexAll {
 
     fn compile(
         &self,
-        state: &state::TypeState,
+        _state: &state::TypeState,
         _ctx: &mut FunctionCompileContext,
         arguments: ArgumentList,
     ) -> Compiled {
         let value = arguments.required("value");
-        let pattern = arguments.required_regex("pattern", state)?;
+        let pattern = arguments.required("pattern");
         let numeric_groups = arguments
             .optional("numeric_groups")
             .unwrap_or_else(|| expr!(false));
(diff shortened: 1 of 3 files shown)


src/stdlib/parse_regex_all.rs
```diff
@@ -105,22 +106,22 @@ impl Function for ParseRegexAll {
...
 
-        parse_regex_all(value, numeric_groups.try_boolean()?, pattern)
+        parse_regex_all(value, numeric_groups.try_boolean()?, &pattern)
     }
 
-    fn type_def(&self, _: &state::TypeState) -> TypeDef {
+    fn type_def(&self, _state: &state::TypeState) -> TypeDef {
         TypeDef::array(Collection::from_unknown(
-            Kind::object(util::regex_kind(&self.pattern)).or_null(),
+            Kind::object(Collection::any()).or_null(),
```
pront, line 124: `self.pattern.resolve_constant()` returns None

spacepatcher:
Hi!
Could you please proceed with considering this small but life-simplifying improvement?

src/stdlib/parse_regex_all.rs
fuchsnj: > self.pattern.resolve_constant(..) returns None
We should make sure at least the case where the regex is passed in as a literal is supported (for backwards compatibility). In that case it should keep using `util::regex_kind`. Let me know if you need help with this, but I thought it would have just worked. It might just be a 1-liner to fix if it doesn't.

For the case where `resolve_constant` returns `None`, you can look at the `regex_kind` function to see what it inserts, and basically build a union of that. It seems like it only every creates an object where the values are bytes or null, but we won't know the key names since they can't be parsed in this situation. So you can create a `Kind::object` with a collection where the unknown type is `bytes | null`.

pront: Thanks, both cases make sense. Updated in https://github.com/vectordotdev/vrl/pull/563/commits/7ffc51f006e36650e057b919b34f8881897399ed. I will follow up with adding a test case (forgot to include in the latest commit).

fuchsnj (approved):

Issue #544, opened 2023-11-03
parse_json gives incorrect results for large integers


djudd:
Example: https://playground.vrl.dev/?state=eyJwcm9ncmFtIjoidmFsdWUgPSBcIjkyMjMzNzIwMzY4NTQ3NzU4MDhcIlxuLnBhcnNlZCwgLmVyciA9IHBhcnNlX2pzb24odmFsdWUpIiwiZXZlbnQiOnt9LCJpc19qc29ubCI6ZmFsc2UsImVycm9yIjpudWxsfQ%3D%3D

input:
```
value = "9223372036854775808"
.parsed, .err = parse_json(value)
```

output:
```
{
	"err": null,
	"parsed": -9223372036854776000
}
```

Some quick experimentation suggests parsing becomes unreliable circa 2^48, but I haven't tested systematically.

(This was found via https://hypothesis.readthedocs.io/en/latest/ tests on my Python code; FWIW I'd suggest using something similar for the parse & encode round-trip in Vector/VRL and might be able to submit that as a contribution if y'all would be interested.)

pront:
Hi @djudd,

Contributions are always welcome. [Here](https://github.com/vectordotdev/vrl/blob/45eb94bd9b427b79f1db4aa9fe0f0bd0ef16b7fe/src/stdlib/parse_json.rs#L10) is the code that parses the JSON bytes. It delegates to `serde_json::from_slice`.

StephenWakely:
This number is greater that is supported by VRL. We should error instead of overflow.

pront:
Agreed, we should go with the least intrusive change i.e. detect that the number cannot be converted to a VRL value and return an error.

Following up on this, @djudd would you like to contribute a fix for this?

djudd:
@pront I'm interested in contributing a fix, but don't want to commit 100% without having given it a shot yet. It'll probably be either next week or not for a while, so if I don't send you a PR by next Friday, please don't block further on me. But I hope to follow up soon.

pront:
Thank you @djudd. If there are no updates by next Friday, I will assign this issue to me.

djudd:
Thank you!

PR #570, opened 2023-11-21
fix!(parse_json): (lossy) convert large numbers to float to avoid int overflow

pront:
closes: #544

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## unreleased
 * `parse_regex_all` `pattern` param  can now be resolved from a variable
+* fixed `parse_json` data corruption issue for numbers greater or equal to `i64::MAX` 
 
 ## `0.8.0` (2023-10-31)
 
--- a/src/stdlib/parse_json.rs
+++ b/src/stdlib/parse_json.rs
@@ -244,9 +244,8 @@ fn type_def() -> TypeDef {
 
 #[cfg(test)]
 mod tests {
-    use crate::value;
-
     use super::*;
+    use crate::value;
 
     test_function![
         parse_json => ParseJson;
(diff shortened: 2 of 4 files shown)


fuchsnj:
This was discussed a bit offline. I realized that if the number is out of bounds of a `u64` (or below the lower bounds of an `i64`) it is already converted to a float today. It would be _really_ odd to have a number return an error here if it's too large, but then using an _even larger_ number is converted to a float.

I understand that converting to a float can lead to precision loss, but I think it's better to keep the float conversion behavior here as well, at least until we have a more comprehensive strategy here that includes the float conversion.

pront:
> This was discussed a bit offline. If the number is out of bounds of a `u64` (or below the lower bounds of an `i64`) it is already converted to a float today. It would be _really_ odd to have a number return an error here if it's too large, but then using an _even larger_ number is converted to a float.
> 
> I understand that converting to a float can lead to precision loss, but I think it's better to keep the float conversion behavior here as well, at least until we have a more comprehensive strategy here that includes the existing float conversion.

Here is an [example](https://playground.vrl.dev/?state=eyJwcm9ncmFtIjoidmFsdWUgPSBcIjkyMjMzNzIwMzY4NTQ3NzU4MDhcIlxuLmEgPSBwYXJzZV9qc29uISh2YWx1ZSlcblxudmFsdWUgPSBcIjkyMjMzNzIwMzY4NTQ3NzU4MDhcIiAqIDJcbi5iID0gcGFyc2VfanNvbiEodmFsdWUpIiwiZXZlbnQiOnt9LCJpc19qc29ubCI6ZmFsc2UsImVycm9yIjpudWxsfQ%3D%3D) demonstrating the current behavior.

Implemented the above suggestion in https://github.com/vectordotdev/vrl/pull/570/commits/118db16a39e17422e7eb8cbc76e3cc89b88b4ef6.

fuchsnj (approved):

PR #573, opened 2023-11-22
feat(compiler): add purity field to TypeDef

pront:


pront:
ref https://datadoghq.atlassian.net/browse/OPW-92

pront:
See https://github.com/vectordotdev/vrl/pull/574

PR #574, opened 2023-11-22
feat(compiler): add purity field to TypeDef

pront:


src/compiler/type_def.rs
```diff
@@ -418,6 +475,7 @@ impl TypeDef {
         Self {
             fallibility: Fallibility::merge(&self.fallibility, &other.fallibility),
             kind,
+            purity: Purity::merge(&self.purity, &other.purity),
```
fuchsnj, line 478: This has me questioning things a bit. I _think_ this is actually correct for purity, but fallibility might have issues. An `insert` can overwrite fields. I believe this function is only used in assignments, and we resolve the RHS side of assignments before the assignment happens, which means side-effects will be ran even if it's later overwritten. However, for fallibility I think the fallibility can actually be passed to the object? But we eventually want to force fallibility checking to happen earlier (before an assignment) so that might not be an issue for long.

This is probably fine for now, just wanted to think out loud here in case that helps anything.

fuchsnj (approved):

src/compiler/type_def.rs
pront: The next PR is going to be interesting, I plan to traverse all expressions (and use the purity information) to determine if an expression result is unused. However, a generalized solution is not trivial. Some examples:

```
# unused result, top level block
parse_json!("1") # (1, false)

. = { 
      { 
        parse_json!("2") # unused result
        "foo"            # unused literal
        parse_json!("4") # used in assignment
      }
}
```

PR #589, opened 2023-12-07
chore: add display for 'PathPrefix'

pront:


fuchsnj (approved):
I'm curious why this was needed / where it's used, but it looks reasonable.

pront:
> I'm curious why this was needed / where it's used, but it looks reasonable.

I was planning to use it in `lib/vector-lookup/src/lookup_v2/mod.rs` but after reading this [comment](https://github.com/vectordotdev/vector/pull/19331#discussion_r1419353938) I realized it is not necessary. Now, it's just a nice to have. I might end up using it for a small refactoring I had in mind in the future.

PR #591, opened 2023-12-08
chore: remove anymap from the dependencies

StephenWakely:
According to https://rustsec.org/advisories/RUSTSEC-2021-0065.html `anymap` is unmaintained.

This PR removes the dependency and replaces it with a simple `HashMap` of `TypeId -> Any`.

### DIFF
--- a/src/compiler/compile_config.rs
+++ b/src/compiler/compile_config.rs
@@ -1,7 +1,12 @@
 use crate::path::OwnedTargetPath;
-use anymap::AnyMap;
-use std::collections::BTreeSet;
+use std::{
+    any::{Any, TypeId},
+    collections::{BTreeSet, HashMap},
+};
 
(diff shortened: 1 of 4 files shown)


src/compiler/compile_config.rs
```diff
@@ -1,7 +1,12 @@
...
-use std::collections::BTreeSet;
+use std::{
+    any::{Any, TypeId},
+    collections::{BTreeSet, HashMap},
+};
 
+type AnyMap = HashMap<TypeId, Box<dyn Any>>;
+
+#[derive(Default)]
 pub struct CompileConfig {
```
pront, line 10: Do we have any unit test coverage for this?

StephenWakely: Good call. I'm not seeing any. I'll add some..

pront (approved):

PR #596, opened 2023-12-12
fix: `LocalEnv::merge` to also merge the `bindings` vector

### LINKED ISSUE #595: How to merge two `TypeState`/`LocalEnv` including the `bindings`?
My use case for VRL is to allow multiple `Program`s to run serially, and I have a shared `RuntimeState` that allows me to share the runtime state. 
I also want to allow variables to be shared across `Program`s. To do that, I'm using `compile_with_state` and passing the previous state. But in the case of >2 `Program`s, I need to merge the `TypeState` of multiple `Program`s. 

```
# Program 1: this one works obviously
my_var1 = "1"

# Program 2: this one also works, because I only need to use a single TypeState (from program 1)
my_var2 = "2"
log("my_var1: " + my_var1 + "my_var2: " + my_var2);

# Program 3: This one fails because "merge" doesn't work as expected
log("my_var1: " + my_var1 + "my_var2: " + my_var2);
```

I tried to use the `LocalEnv::merge` function, but it seems like it's merging the `details` only, without merging the `bindings` vector together:
https://github.com/vectordotdev/vrl/blob/59fb310809ac7bb9f2e1b137d10ca4178a4e5625/src/compiler/state.rs#L87-L93

I suspect that `LocalEnv::merge` might have a bug because it does not merge the complete `bindings` vectors. I tried to work around it but it seems like all functions and properties are `pub(crate)` so I can't seem to create the `LocalEnv` manually with the data I need. 
I created a PR to address this issue: https://github.com/vectordotdev/vrl/pull/596

dotansimha:
Closes https://github.com/vectordotdev/vrl/issues/595

### DIFF
--- a/src/compiler/state.rs
+++ b/src/compiler/state.rs
@@ -88,6 +88,8 @@ impl LocalEnv {
         for (ident, other_details) in other.bindings {
             if let Some(self_details) = self.bindings.get_mut(&ident) {
                 *self_details = self_details.clone().merge(other_details);
+            } else {
+                self.bindings.insert(ident, other_details);
             }
         }
         self
--- a/src/compiler/target.rs
+++ b/src/compiler/target.rs
@@ -278,7 +278,10 @@ mod value_target_impl {
 mod tests {
     #![allow(clippy::print_stdout)] // tests
 
-    use crate::owned_value_path;
+    use crate::{
+        compiler::{parser::Ident, state::LocalEnv, type_def::Details, TypeDef},
+        owned_value_path,
+    };
 
     use super::*;
     use crate::value;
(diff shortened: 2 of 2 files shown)


pront (commented):
Looks reasonable. Left one comment.

dotansimha:
thanks @pront ! I don't see the comment you mentioned 👀

src/compiler/state.rs
```diff
@@ -88,6 +88,8 @@ impl LocalEnv {
         for (ident, other_details) in other.bindings {
             if let Some(self_details) = self.bindings.get_mut(&ident) {
                 *self_details = self_details.clone().merge(other_details);
+            } else {
```
pront, line 91: - How did you test this?
- Can you add a #[test] please?

pront (approved):
LGTM

PR #622, opened 2024-01-04
feat(diagnostics): unused expression result detector

pront:
https://datadoghq.atlassian.net/browse/OPW-92

## Summary
* After compilation, do another AST traversal
* Discover unused expression results
  * including unused variables (shadowing not supported)
* Produce warnings including source code spans for the unused expression
* The goal is to catch as much unused expressions as possible but not all of them. We do have to eliminate all false positives though.

## Testing plan
  * Unit tests
  * Tweak the testing harness to compare emitted diagnostics
  * Fix all existing VRL tests (warnings are treated as errors by the test harness)


src/compiler/unused_expression_checker.rs
```diff
@@ -0,0 +1,680 @@
...
+    state.mark_level_as_expecting_result();
+    f(state);
+    state.mark_level_as_not_expecting_result();
+    state.decrease_level();
+}
+
+impl AstVisitor<'_> {
+    fn visit_node(&self, node: &Node<Expr>, state: &mut VisitorState) {
+        let expression = node.inner();
+        // println!("\n{} visit_node: {expression:#?}", state.level);
```
pront, line 175: Reminder: All `// println!(...` lines should be be deleted before merging.

fuchsnj (approved):
This looks reasonable from a high level. I'm glad the `unused_expression_checker` was kept separate from the main compiler, that should hopefully keep it more maintainable.

Issue #628, opened 2024-01-07
regex replace using a closure


tmccombs:
I filed https://github.com/vectordotdev/vrl/issues/633, but now realize that this might have been a better place to report it. 

For convenience I will copy over my text from there:

### Use Cases

I want to redact certain personally identifiable information, such as an email address, from logs, but in such a way that if I already know the information (email address), I can search the logs for matching log entries. 

### Attempted Solutions

I think it would be possible to do this with a lua transform. But  the documentation for that says to create an issue if the remap transform doesn't meet my needs.

Lua also doesn't have a built in sha256 function so I would either need to use a lua native sha256 implementation, which I suspect would be slow, or pull in a shared library that adds such a function for lua (is that possible with vector?)

### Proposal

I can think of a few ways this could be done:

- Add an option to the redact function that will replace the matched string with a hash of itself, Possibly with a configurable hash function, instead of with a fixed string.
- add a new function  that replaces a matching regex with its hash
- make a function like replace, but instead of a string for the replacement, it uses a closure to compute a replacement value.

I think that the last bullet point is the most general and could be useful in other cases as well. 

P.s. having a built in redaction pattern for email addresses would be awesome.

tmccombs:
I don't currently have time to put together a full PR, but here is a rough draft of a partial implementation of a replace function that uses a closure:

```rust
fn replace_with<T>(value: Value, pattern: Value, ctx: &mut Context, runner: closure::Runner<T>) -> Resolved {
    let value = value.try_bytes_utf8_lossy()?;
    // TODO: support count?
    match pattern {
        Value::Regex(regex) => {
            let mut i: usize = 0;
            let mut failure: Option<ExpressionError> = None;

            let replaced = regex.replace_all(&value, |captures| {
                let captures_value = captures_to_value(captures);
                let result = runner.run_index_value(ctx, i, captures_value).and_then(|s| Ok(s.try_bytes_utf8_lossy()?));
                match result {
                    Ok(v) => v,
                    Err(e) => {
                        failure.get_or_insert(e);
                        Cow::from("")
                    }
                }
            }).as_bytes();
            if let Some(err) = failure {
                Err(e)
            } else {
                Ok(replaced.into())
            }
        }
        // TODO: should we also support Value::Bytes?
        value => Err(ValueError::Expected {
            got: value.kind(),
            expected: Kinde.regex()
        }.into())
    }
}

fn coaptures_to_value(captures: &Captures) -> Value {
[...]

PR #636, opened 2024-01-11
feat(stdlib): Add replace_with function

tmccombs:
This is similar to `replace`, but takes a closure to compute the replacment from the match and capture groups, instead of taking a replacment string.

Fixes: #628

### DIFF
--- a/src/stdlib/replace_with.rs
+++ b/src/stdlib/replace_with.rs
@@ -3,7 +3,6 @@ use std::collections::BTreeMap;
 use regex::{CaptureMatches, CaptureNames, Captures, Regex};
 
 use crate::compiler::prelude::*;
-use crate::value;
 
 fn replace_with<T>(
     value: Value,
@@ -70,11 +69,16 @@ where
     Ok(replaced.into())
 }
 
+const STRING_NAME: &str = "string";
+const CAPTURES_NAME: &str = "captures";
+
 fn captures_to_value(captures: &Captures, capture_names: CaptureNames) -> Value {
-    let full_match: Value = captures.get(0).unwrap().as_str().into();
+    let mut object: ObjectMap = BTreeMap::new();
+
+    // The full match, named "string"
+    object.insert(STRING_NAME.into(), captures.get(0).unwrap().as_str().into());
     // The length includes the total match, so subtract 1
     let mut capture_groups: Vec<Value> = Vec::with_capacity(captures.len() - 1);
-    let mut named_groups: ObjectMap = BTreeMap::new();
 
     // We skip the first entry, because it is for the full match, which we have already
     // extracted
@@ -85,16 +89,14 @@ fn captures_to_value(captures: &Captures, capture_names: CaptureNames) -> Value 
             Value::Null
         };
         if let Some(name) = name {
-            named_groups.insert(name.into(), value.clone());
+            object.insert(name.into(), value.clone());
         }
         capture_groups.push(value);
     }
 
-    value!({
-        string: full_match,
-        captures: capture_groups,
-        named: named_groups,
-    })
+    object.insert(CAPTURES_NAME.into(), capture_groups.into());
+
+    object.into()
 }
 
 #[derive(Clone, Copy, Debug)]
@@ -154,7 +156,7 @@ impl Function for ReplaceWith {
             },
             Example {
                 title: "Named capture group",
-                source: r#"replace_with("foo123bar", r'foo(?P<num>\d+)bar') -> |m| { x = to_int!(m.named.num); to_string(x+ 1) }"#, //to_string(to_int!(m.named.num) + 1) }"#,
+                source: r#"replace_with("foo123bar", r'foo(?P<num>\d+)bar') -> |m| { x = to_int!(m.num); to_string(x+ 1) }"#, //to_string(to_int!(m.named.num) + 1) }"#,
                 result: Ok("\"124\""),
             },
         ]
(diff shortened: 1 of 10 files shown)


src/stdlib/replace_with.rs
```diff
@@ -0,0 +1,214 @@
...
+        _ctx: &mut FunctionCompileContext,
+        arguments: ArgumentList,
+    ) -> Compiled {
+        let value = arguments.required("value");
+        let pattern = arguments.required("pattern");
+        let count = arguments.optional("count").unwrap_or(expr!(-1));
+
+        let closure = arguments.required_closure()?;
+
+        // TODO: Should I check that the closure is not fallible?
```
pront, line 152: Let's add some VRL tests (if unit tests for closures are a challenge). See this for example: https://github.com/vectordotdev/vrl/blob/main/lib/tests/tests/expressions/function_call/closure_scope_inheritance.vrl. You can just add ".vrl" files under `lib/tests/tests/*` and they will be discovered. Note that examples will also run as tests.

src/stdlib/replace_with.rs
```diff
@@ -0,0 +1,214 @@
...
+                keyword: "count",
+                kind: kind::INTEGER,
+                required: false,
+            },
+        ]
+    }
+
+    fn examples(&self) -> &'static [Example] {
+        &[
+            Example {
```
pront, line 112: We will also need a PR in the Vector repo because that's where we generate the public docs (hopefully that can be simplified in the future). See [this for example](https://github.com/vectordotdev/vector/blob/master/website/cue/reference/remap/functions/replace.cue).

pront (commented):
Hi @tmccombs and thank you for this PR! Mostly looks good, left a couple of comments.

src/stdlib/replace_with.rs
```diff
@@ -0,0 +1,214 @@
...
+        if i >= limit {
+            break;
+        }
+    }
+    // add the final component
+    replaced.push_str(&haystack[last_match..]);
+    Ok(replaced.into())
+}
+
+fn captures_to_value(captures: &regex::Captures) -> Value {
```
tmccombs, line 71: Is using an array the best representation for the match? would it be better to use an object with  a key for the whole match, a key fro the capture groups, and maybe something for the named capture groups?

src/stdlib/replace_with.rs
tmccombs: I created https://github.com/vectordotdev/vector/pull/19638

src/stdlib/replace_with.rs
tmccombs: I've added some.

src/stdlib/replace_with.rs
pront: Thanks! I will review/merge that after the next VRL release.

src/stdlib/replace_with.rs
pront: Regarding the TODO, I think we can remove it.

src/stdlib/replace_with.rs
```diff
@@ -0,0 +1,214 @@
...
+    value: Box<dyn Expression>,
+    pattern: Box<dyn Expression>,
+    count: Box<dyn Expression>,
+    closure: FunctionClosure,
+}
+
+impl FunctionExpression for ReplaceWithFn {
+    fn resolve(&self, ctx: &mut Context) -> ExpressionResult<Value> {
+        let value = self.value.resolve(ctx)?;
+        let pattern = self.pattern.resolve(ctx)?;
```
pront, line 200: For resolving the pattern, see this example: https://github.com/vectordotdev/vrl/pull/563/files#diff-8f4955c315c501ce024ba7c200c8af2a9aa33c59db955c305b99615690c83e9dR117-R122

src/stdlib/replace_with.rs
tmccombs: removed

src/stdlib/replace_with.rs
```diff
@@ -0,0 +1,212 @@
...
+where
+    T: Fn(&mut Context) -> Result<Value, ExpressionError>,
+{
+    // possible optimization: peek at first capture, if none return the original value.
+    let mut replaced = String::with_capacity(haystack.len());
+    let limit = if count == 0 { usize::MAX } else { count - 1 };
+    let mut last_match = 0;
+    // we loop over the matches ourselves instead of calling Regex::replacen, so that we can
+    // handle errors. This is however based on the implementation of Regex::replacen
+    for (i, cap) in caps.enumerate() {
```
pront, line 51: Nit: Let's improve the variable name, basically to avoid abbreviations

src/stdlib/replace_with.rs
pront: Perhaps an object would be cleaner. Up to you.

tmccombs: I decided to go with an object, so that we can expose the named capture groups by name. 

The downsides of this approach are:
- If there are named capture groups, we have to clone the strings (I wonder if an `Arc<str>` instead of a string for a string value would benefit performance)
- It's a little more verbose to reference non-named capture groups

src/stdlib/replace_with.rs
```diff
@@ -0,0 +1,248 @@
...
+        if let Some(name) = name {
+            named_groups.insert(name.into(), value.clone());
+        }
+        capture_groups.push(value);
+    }
+
+    value!({
+        string: full_match,
+        captures: capture_groups,
+        named: named_groups,
```
tmccombs, line 96: I wonder if it would be better to put the named capture groups directly on the the returned object.  But then what do we do if there is a capture group named "string" or "captures"?

pront: This sounds cleaner. Not sure I understand the question. Can you give an example?

tmccombs: If the regex is something like `r'"(?P<string>.*)"''` should I:

1. Set the `string` property to the capture group for "string"?
2. Rename the capture group to something else like "string_"
3. Not use a named property for the capture group
4. Return an error because there is an illegal name

pront: Ah makes sense, thanks for the clear explanation!

Let's go for 4.

pront (approved):

pront:
Thanks @tmccombs, I will merge after you fix the fmt and clippy checks.

tmccombs:
fixed

PR #640, opened 2024-01-12
enhancement(redact): Add redactor option

tmccombs:
This option allows you to specify either a custom string, or a hash function to use as the redactor, in addition to the current behavior of the fixed string "[Redacted]".


src/stdlib/redact.rs
```diff
@@ -220,52 +241,162 @@ impl Filter {
...
+                _ => Err("unknown name of redactor"),
+            },
+            _ => Err("unknown literal for redactor, must be redactor name or object"),
         }
     }
 }
 
+/// Compute the digest of some bytes as hex string
+fn hex_digest<T: digest::Digest>(value: &[u8]) -> String {
+    hex::encode(T::digest(value))
```
tmccombs, line 397: would it be better to use base64 encoding? or have another option for the text encoding so the caller can choose between hex and base64?

jszwedko: As a potential user of this, what would you find most useful? My instinct is that base64 would be more common. Maybe we could use the same approach as we use for the `filters` argument to this function where additional options could be specified, if needed, for a given redactor like:

```
redact("my id is 123456", filters: [r'\d+'], redactor: {name: "sha256", encoding: "base64"})
```

But where you could also just specify:

```
redact("my id is 123456", filters: [r'\d+'], redactor: "sha256")
```

For the default behavior (say base64 encoding). What would you think?

tmccombs: > My instinct is that base64 would be more common

that is my instinct as well.

src/stdlib/redact.rs
```diff
@@ -53,6 +58,30 @@ impl Function for Redact {
...
             },
+            Example {
+                title: "text redactor",
+                source: r#"redact("my id is 123456", filters: [r'\d+'], redactor: {"type": "text", "replacement": "***"})"#,
+                result: Ok(r#"my id is ***"#),
+            },
+            Example {
+                title: "sha2",
+                source: r#"redact("my id is 123456", filters: [r'\d+'], redactor: "sha2")"#,
+                result: Ok(r#"my id is GEtTedW1p6tC094dDKH+3B8P+xSnZz69AmpjaXRd63I="#),
```
pront, line 69: 🥜 Would it better if we wrapped the result e.g. 
```
Ok(r#"my id is `GEtTedW1p6tC094dDKH+3B8P+xSnZz69AmpjaXRd63I=`"#),
```

pront (commented):
Overall this looks good!

- Can you also update the vector docs (see [this example](https://github.com/vectordotdev/vector/pull/19638)) please?
- Examples look great. Can you add similar unit tests?

src/stdlib/redact.rs
tmccombs: maybe? I don't have a preference either way.

pront: In this case, let's leave it as is

tmccombs:
vector documentation PR: https://github.com/vectordotdev/vector/pull/19749

pront:
> vector documentation PR: [vectordotdev/vector#19749](https://github.com/vectordotdev/vector/pull/19749)

Awesome, thank you.

changelog.d/633.enhancement.md
```diff
@@ -0,0 +1,2 @@
+Added `redactor` option to `redact` to allow replacing with a custom string or hash of the redacted
+content.
```
pront, line 2:
```suggestion
Introduced a redactor option in redact to enable the substitution of redacted content with either a custom string or a hash representation.
```

pront (approved):
Added one suggestion. Otherwise, this looks good. Thank you!

pront:
@tmccombs this looks good to me. Do you think it is ready to merge?

jszwedko:
@tmccombs it looks like there are some merge conflicts now too when you get a second.

tmccombs:
yeah, I think this is ready to merge.

PR #649, opened 2024-01-22
chore(scripts): update publish script to also generate the new changelog

pront:


pront:
Closing this, it doesn't make sense to generate at this point. I might add further validation though (check if there are remaining fragments) in a follow up PR.

PR #652, opened 2024-01-23
chore: update publish script to check if there are fragments

pront:
The purpose of this PR is to add an extra sanity check before publishing a new VRL version.

### DIFF
--- a/scripts/publish.py
+++ b/scripts/publish.py
@@ -8,7 +8,7 @@ import toml
 
 SCRIPTS_DIR = os.path.dirname(abspath(getsourcefile(lambda: 0)))
 REPO_ROOT_DIR = os.path.dirname(SCRIPTS_DIR)
-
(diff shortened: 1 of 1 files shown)


scripts/publish.py
```diff
@@ -38,19 +38,31 @@ def publish_vrl(version):
         print(f"An error occurred while publishing the crate: {e}")
 
 
-def main():
+def assert_no_changelog_fragments():
+    entries = os.listdir(CHANGELOG_DIR)
+    error = f"{CHANGELOG_DIR} should only contain a README.md file. Did you run ./scripts/generate_changelog.py?"
```
neuronull, line 43:
```suggestion
    error = f"{CHANGELOG_DIR} should only contain a README.md file. Did you run ./scripts/generate_release_changelog.sh?"
```

neuronull (approved):

Issue #659, opened 2024-01-26
Feature request: detect, parse punycode for DNS IDN objects


johnhtodd:
Any system that digests DNS data for more than a few hostnames will eventually get IDN names (https://unicode.org/faq/idn.html) in the data stream. This quickly presents problems for back-end data store and monitoring systems.

Systems may choose to keep IDN in their existing form, but having the ability to add fields which are encoded in punycode (https://en.wikipedia.org/wiki/Punycode) is desirable, as many DNS logging systems expect and use punycode to simplify management of handling IDN objects.

There are two three things which I think would be useful to have in Vector:
 - the ability to detect if a field contains characters which are present in IDN-formatted strings
 - the ability to convert IDN names into punycode from UTF-8
 - the ability to convert IDN names from punycode into UTF-8

There appear to be several existing Rust libraries which do this type of conversion. I don't know which have what desirable features or preferred licenses, so I won't make references here so as not to bias any search.

Notes: 
  - this is DNS-specific.  The conversion would have to be applied to each "zone cut" between periods, but portions of the name that do not have UTF-8 would not be translated.  So www.café.com would be www.xn--caf-dma.com as an example.
  -  lowercasing the name is automatically assumed.  So www.CAFé.com and www.cafÉ.com would also be www.xn--caf-dma.com

jszwedko:
Thanks for opening this @johnhtodd ! I agree this would make sense as a VRL function.

PR #672, opened 2024-02-05
feat(stdlib): add punycode encoding functions

esensar:
This adds `encode_punycode` and `decode_punycode` functions. It also adds tests to confirm `parse_url` function behavior when it comes to punycode.

Fixes: #659


esensar:
This adds punycode related functions. I have also realised that `parse_url` already handles that, so I have added a few tests to confirm it. The newly added crate for handling punycode was already a transitive dependency of the project (because of the use of `url` crate, and `idna` crate which handles punycode is a direct dependency of `url` and maintained in the same repository https://github.com/servo/rust-url).

johnhtodd:
I haven't tested this patch, so forgive me if this lazy request could be answered that way: What happens if a string not containing UTF-8 characters is run through the encode_punycode process?  Or a non-punycode string is run through decode_punycode?

esensar:
Right, I should have added some tests and examples to demonstrate that behavior. To answer your question, it will only encode whatever requires encoding, so for that reason the example (www.café.com) only encodes the café part.

I will add tests and examples with fully ASCII domains to make this more clear.

johnhtodd:
Thanks - more test cases are always good.  Very happy to see this feature.  I suspect it will have a better chance of approval as a PR if it also had documentation.

pront (commented):
Thank you @esensar, this looks good. 

Can you also add a VRL test [like this](https://github.com/vectordotdev/vrl/blob/main/lib/tests/tests/functions/parse_groks.vrl) that exercises both encode and decode?

jszwedko:
> Thank you @esensar, this looks good.
> 
> Can you also add a VRL test [like this](https://github.com/vectordotdev/vrl/blob/main/lib/tests/tests/functions/parse_groks.vrl) that exercises both encode and decode?

Additionally, could we add these new functions to the benchmarks in `benches/stdlib.rs`?

esensar:
> > Thank you @esensar, this looks good.
> > Can you also add a VRL test [like this](https://github.com/vectordotdev/vrl/blob/main/lib/tests/tests/functions/parse_groks.vrl) that exercises both encode and decode?
> 
> Additionally, could we add these new functions to the benchmarks in `benches/stdlib.rs`?

Yeah, I have added them now.

I have added both cases when there is nothing to encode/decode and with IDN strings, to confirm that cases when no encoding is needed are faster (an order of magnitude faster on my machine).

esensar:
I have added VRL tests as well, demonstrating encoding, decoding and combination with `parse_url` as well. I realized that I have made the function infallible accidentally, so I have updated that too.

lib/tests/tests/functions/punycode/encode_decode.vrl
```diff
@@ -0,0 +1,5 @@
+# object: { "host": "www.ليهمابتكلموشعربي؟.他们为什么不说中文" }
+# result: "www.ليهمابتكلموشعربي؟.他们为什么不说中文"
+
+encoded = encode_punycode!(.host)
+decode_punycode!(encoded)
```
pront, line 5: Thanks for adding this, looks good.

src/stdlib/decode_punycode.rs
```diff
@@ -0,0 +1,87 @@
...
+    value: Box<dyn Expression>,
+}
+
+impl FunctionExpression for DecodePunycodeFn {
+    fn resolve(&self, ctx: &mut Context) -> Resolved {
+        let value = self.value.resolve(ctx)?;
+        let string = value.try_bytes_utf8_lossy()?;
+
+        let (encoded, result) = idna::domain_to_unicode(&string);
+        result.map_err(|err| format!("unable to decode punycode: {err}"))?;
```
pront, line 57: 🥜 rename `err` to `errors`

src/stdlib/encode_punycode.rs
```diff
@@ -0,0 +1,98 @@
...
+    value: Box<dyn Expression>,
+}
+
+impl FunctionExpression for EncodePunycodeFn {
+    fn resolve(&self, ctx: &mut Context) -> Resolved {
+        let value = self.value.resolve(ctx)?;
+        let string = value.try_bytes_utf8_lossy()?;
+
+        let encoded = idna::domain_to_ascii(&string)
+            .map_err(|err| format!("unable to encode to punycode: {err}"))?;
```
pront, line 62: 🥜 rename `err` to `errors`

pront (approved):
Thank you for adding the tests!

PR #679, opened 2024-02-06
Fix two bugs with exporting paths

bruceg:
This series fixes handling exporting owned paths containing more than one coalesce segment, stemming from a mutable variable not being reset in a loop.


pront (approved):
Neat, can you share how ran the `proptest`s?

bruceg:
> Neat, can you share how ran the `proptest`s?

For sure, that's coming up next.

PR #680, opened 2024-02-06
Run proptests on paths

bruceg:
Add property tests to ensure that arbitrary path values can be exported to strings and then re-parsed from that string to produce an equal path. This required some tweaks to the automatically generated `Arbitrary` implementations, and I added implementations of `FromStr` along the way to simplify parsing.

### DIFF
--- a/src/path/mod.rs
+++ b/src/path/mod.rs
@@ -311,7 +311,7 @@ fn get_target_prefix(path: &str) -> (PathPrefix, &str) {
 }
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
-#[cfg_attr(feature = "proptest", derive(proptest_derive::Arbitrary))]
+#[cfg_attr(any(test, feature = "proptest"), derive(proptest_derive::Arbitrary))]
 pub enum PathPrefix {
     Event,
     Metadata,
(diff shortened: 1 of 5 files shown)


src/path/owned.rs
```diff
@@ -547,4 +597,25 @@ mod test {
             assert_eq!(path, expected.map(|x| x.to_owned()));
         }
     }
+
+    fn reparse_thing<T: std::fmt::Debug + std::fmt::Display + Eq + FromStr>(thing: T)
```
pront, line 601: does this need a `proptest` feature gate?

src/path/mod.rs
```diff
@@ -311,7 +311,7 @@ fn get_target_prefix(path: &str) -> (PathPrefix, &str) {
 }
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
-#[cfg_attr(feature = "proptest", derive(proptest_derive::Arbitrary))]
+#[cfg_attr(any(test, feature = "proptest"), derive(proptest_derive::Arbitrary))]
```
pront, line 314: Do we want `all` vs `any`?

src/path/owned.rs
bruceg: No, because proptests are always enabled for testing.

src/path/mod.rs
bruceg: As above, I enabled proptests for testing, so these features are enabled for both `cargo test` and when the `proptest` feature is enabled by another crate (i.e. Vector).

pront (approved):

PR #681, opened 2024-02-06
Optimize path export functions

bruceg:
This is a series of incremental optimizations to the handling of exporting owned paths to strings. Mostly they are a process of eliminating intermediate `String` values and implementing the string conversion in terms of the `Display` implementations instead of the other way around.

FWIW I also did `Display` conversion for `OwnedValuePath` itself, but the underlying `serialize_field` function ended up with a looped `write!` of single characters, which seemed unlikely to be performant. Fixing that is probably more work than it's worth.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -126,7 +126,7 @@ hmac = { version = "0.12.1", optional = true }
 iana-time-zone = "0.1.59"
 indexmap = { version = "~2.2.2", default-features = false, features = ["std"], optional = true}
 indoc = {version = "2.0.4", optional = true }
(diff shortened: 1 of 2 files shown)


pront:
Thank you, this is neat. 

- How did you test this? 
- Can you also run the Vector tests using this PR as the VRL dep version?

bruceg:
It is tested using the new proptests, which exercises all of the code except for `Display` on `OwnedSegment`, and the internal VRL tests, which exercises that excepted bit. I can definitely pull it into Vector to test if that will give more confidence.

bruceg:
The Vector VRL tests pass as well.

pront (approved):

PR #696, opened 2024-02-13
Make iana-time-zone optional and behind stdlib flag

klondikedragon:
As requested here: https://github.com/vectordotdev/vrl/pull/671#discussion_r1479769824

Please add the no-changelog label to the PR.

Tested to pass the checks script.

Note that iana-time-zone is a dependency of the chrono crate, which is a dependency of the compiler and value features, so this crate is usually pulled in anyway, but this is definitely cleaner.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,6 +74,7 @@ stdlib = [
  "dep:hex",
  "dep:hmac",
  "dep:hostname",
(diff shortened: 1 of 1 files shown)


jszwedko:
Thanks @klondikedragon !

pront (approved):

PR #709, opened 2024-02-21
feat(stdlib): add option to skip validation for punycode functions

esensar:
Adds additional optional argument to `encode_punycode` and `decode_punycode` that disables validation. This is useful in cases when it is not important if the result is a valid domain name. By default validation is enabled.

### DIFF
--- a/changelog.d/709.feature.md
+++ b/changelog.d/709.feature.md
@@ -1,0 +1,2 @@
+Added `validate` option to `encode_punycode` and `decode_punycode`, which defaults to true, but can
+be used to skip validation when set to false.
--- a/src/stdlib/decode_punycode.rs
+++ b/src/stdlib/decode_punycode.rs
@@ -9,11 +9,18 @@ impl Function for DecodePunycode {
     }
 
     fn parameters(&self) -> &'static [Parameter] {
-        &[Parameter {
-            keyword: "value",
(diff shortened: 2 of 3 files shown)


jszwedko (commented):
Given the default of `validate` is `true`, is this a breaking change to behavior? If so, we may want to default it to `false`.

esensar:
> Given the default of `validate` is `true`, is this a breaking change to behavior? If so, we may want to default it to `false`.

Current behavior is to always validate, so this just adds a way to skip validation, if one does not care if it is a valid domain name.

pront (approved):
Let's also update the Vector-VRL docs and explain why this is a useful argument.

esensar:
> Let's also update the Vector-VRL docs and explain why this is a useful argument.

I have added the docs here: https://github.com/vectordotdev/vector/pull/19923
Let me know if more details are required.

jszwedko:
> > Given the default of `validate` is `true`, is this a breaking change to behavior? If so, we may want to default it to `false`.
> 
> Current behavior is to always validate, so this just adds a way to skip validation, if one does not care if it is a valid domain name. So this should not be a breaking change.

Aha, gotcha 👍 then since this is backwards compatible.

jszwedko (approved):

PR #710, opened 2024-02-22
fix(stdlib): ensure punycode functions keep all input parts

esensar:
I have implemented validation skip in #709, which ignores errors returned by `idna` crate when encoding and decoding punycode. One of the errors that it could return would result in output with missing sections (as was demonstrated in tests - e.g. `multiple_errors_ignore`). Since it is fairly trivial to split domain into parts (`idna` crate does it by doing split on `.`), we can split the domain ourselves in cases when we don't want validation and then try to encode ones that need encoding and leave them as they were if encoding (or decoding) fails.

Related: #709

### DIFF
--- a/src/stdlib/decode_punycode.rs
+++ b/src/stdlib/decode_punycode.rs
@@ -1,5 +1,7 @@
 use crate::compiler::prelude::*;
 
+const PUNYCODE_PREFIX: &str = "xn--";
+
 #[derive(Clone, Copy, Debug)]
 pub struct DecodePunycode;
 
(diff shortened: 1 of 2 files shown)


pront (approved):

esensar:
I haven't added a changelog since it affects changes that were already covered by a changelog entry that wasn't yet released. Is that alright or should I add one?

jszwedko:
> I haven't added a changelog since it affects changes that were already covered by a changelog entry that wasn't yet released. Is that alright or should I add one?

👍 that seems right since the other change hasn't been released yet.

Issue #704, opened 2024-02-19
Add "sieve" string function


johnhtodd:
It would be useful to have a method to remove characters from strings which are undesired or un-necessary for parsing of data (typically distilling to human-readable data fields.)  This would eliminate or replace characters that have special (dangerous) meaning to downstream consumers, or which are not presentable in other formats. 

Purposes for this function are multiple: it provides an additional layer of safety for downstream systems which may not handle characters appropriately; it may reduce the volume of data where there are large numbers of unprintable or unstructured characters in a message; it may make more readable certain messages that contain excessive numbers of extraneous characters.

I would propose the function name of "sieve" for this function, and at first glance I would imagine three arguments:

  - string to act upon (required)
  - set of characters to permit, in example form:   [A-z],[0-9], \,, ü, 0xE282AC  (required) 
     - this would be ranges of ASCII or UTF-8 
     - actual characters embedded in config file (is this legal in Vector configs?)
     - hexadecimal UTF-8 
     - escaped (\) special characters
  - character to insert/overwrite single instances of deleted results in output (default is to delete all characters other than sieve list.) Can be explicitly set to "" to delete characters with no replacement. (optional)
  - character to insert/overwrite when two or more sequential instances of deleted characters are in input string. Can be explicitly set to "" to delete characters with no replacement. (optional)

PR #724, opened 2024-03-04
feat(stdlib): add `sieve` string function

esensar:
Adds a `sieve` string function which can remove unwanted characters from a string using a list of allowed characters (or regex of allowed patterns).

Fixes: #704

### DIFF
--- a/src/stdlib/sieve.rs
+++ b/src/stdlib/sieve.rs
@@ -1,5 +1,3 @@
-use std::collections::HashSet;
-
 use crate::compiler::prelude::*;
 
 fn sieve(
@@ -13,26 +11,6 @@ fn sieve(
     let replace_repeated = replace_repeated.try_bytes_utf8_lossy()?;
 
     match permitted_characters {
-        Value::Bytes(bytes) => {
-            let string = String::from_utf8_lossy(&bytes);
-            let characters: HashSet<char> = string.chars().collect();
-            let mut result = String::with_capacity(value.len());
-            let mut missed_length = 0;
-            for char in value.chars() {
-                if characters.contains(&char) {
-                    match missed_length {
-                        l if l > 1 => result += &replace_repeated,
-                        1 => result += &replace_single,
-                        _ => (),
-                    }
-                    missed_length = 0;
-                    result.push(char);
-                } else {
-                    missed_length += 1;
-                }
-            }
-            Ok(result.into())
-        }
         Value::Regex(regex) => {
             let mut result = String::with_capacity(value.len());
             let mut last_end = 0;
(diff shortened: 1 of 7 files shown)


src/stdlib/sieve.rs
```diff
@@ -0,0 +1,196 @@
...
+                kind: kind::BYTES | kind::REGEX,
+                required: true,
+            },
+            Parameter {
+                keyword: "replace_single",
+                kind: kind::BYTES,
+                required: false,
+            },
+            Parameter {
+                keyword: "replace_repeated",
```
pront, line 84: ❔ Can you explain a bit what this `replace_repeated` argument does? I think we could come up with a better name.

esensar: Yeah, it is a bit confusing.

Based on the original issue (https://github.com/vectordotdev/vrl/issues/704), it is meant to provide a way to set a replacement string when multiple repeated characters are removed from the original string.

You can also see how it behaves in tests and examples.

`all_options` example might represent this the best:
```rust
all_options {
    args: func_args![value: value!("test123%456.فوائد.net."), permitted_characters: regex::Regex::new("[a-z.0-9]").unwrap(), replace_single: "X", replace_repeated: "<REMOVED>"],
    want: Ok(value!("test123X456.<REMOVED>.net.")),
    tdef: TypeDef::bytes().infallible(),
}
```

In this example the single `%` found in the string that was not allowed, was replaced with `X`, but that Arabic (maybe?) script string was replaced with `<REMOVED>`.

src/stdlib/sieve.rs
```diff
@@ -0,0 +1,196 @@
...
+    let replace_repeated = replace_repeated.try_bytes_utf8_lossy()?;
+
+    match permitted_characters {
+        Value::Bytes(bytes) => {
+            let string = String::from_utf8_lossy(&bytes);
+            let characters: HashSet<char> = string.chars().collect();
+            let mut result = String::with_capacity(value.len());
+            let mut missed_length = 0;
+            for char in value.chars() {
+                if characters.contains(&char) {
```
pront, line 22: Is there any crate that uses a more efficient algorithm for string matching?

esensar: I will check it out and let you know.

pront: Btw, this is an optional step, I was curious. If there's nothing mature available we can just leave a comment that this can be optimized.

esensar: Sorry, I forgot to respond. I was thinking, maybe it would make sense to remove the option to use `string` and enforce only `regex`. On that one at least no manual checks are performed and I just iterate through the returned matches.

pront: Sounds reasonable to me

pront (approved):
Thank you for this contribution!

PR #729, opened 2024-03-06
chore(deps): Relax version dependency constraints

jszwedko:
To allow for semantically versioned dependency equivalents. When trying to downgrade flate2 in Vector for testing it complained that the VRL crate required a newever version of flate2 since 1.0.28 restricts to >=1.0.28 and < 2.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -103,59 +103,59 @@ stdlib = [
  ]
 
 [dependencies]
-cfg-if = "1.0.0"
+cfg-if = "1"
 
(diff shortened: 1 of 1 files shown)


pront (approved):
Thank you for updating these. I am curious, were all these necessary for the flate2 version change?

jszwedko:
> Thank you for updating these. I am curious, were all these necessary for the flate2 version change?

Nah, just the `flate2` one was needed. I just figured I'd update the rest while I was at it.

PR #735, opened 2024-03-08
chore: Set user-agent for publish script

jszwedko:
crates.io has started returning 403s for the default requests user-agent. I'm guessing to avoid bot requests.


pront (approved):
👍

PR #736, opened 2024-03-10
fix(parse_key_value): handle more escapes (#725)

### LINKED ISSUE #725: parse_logfmt adds extra fields
### A note for the community

### Problem

I use parse_logfmt and I get field which I didn't expect to get.

in the following link you can see **"\\t\\t\"": true** in the output. This field should be part of errorVerbose value, it should be standalone field.
https://playground.vrl.dev/?state=eyJwcm9ncmFtIjoibG9nLCBlcnIgID0gIHBhcnNlX2xvZ2ZtdCgubWVzc2FnZSlcbi4gPSBtZXJnZSguLCBsb2cpIiwiZXZlbnQiOnsibWVzc2FnZSI6ImxldmVsPWVycm9yIGVycm9yVmVyYm9zZT1cImNhbm5vdCBcXHRcXHRcIiJ9LCJpc19qc29ubCI6ZmFsc2UsImVycm9yIjpudWxsfQ%3D%3D

but if I take one '\' off I get a good output, without extra fields
https://playground.vrl.dev/?state=eyJwcm9ncmFtIjoibG9nLCBlcnIgID0gIHBhcnNlX2xvZ2ZtdCgubWVzc2FnZSlcbi4gPSBtZXJnZSguLCBsb2cpIiwiZXZlbnQiOnsibWVzc2FnZSI6ImxldmVsPWVycm9yIGVycm9yVmVyYm9zZT1cImNhbm5vdCBcdFxcdFwiIn0sImlzX2pzb25sIjpmYWxzZSwiZXJyb3IiOm51bGx9

I would expect to get errorVerbose value as one line value, without being split, for example:
{
	"errorVerbose": "cannot \\t\\t",
	"level": "error",
	"message": "level=error errorVerbose=\"cannot \\t\\t\""
}

I can see the extra fields is added when there is \\x\\y pattern inside the value in the input (x is a single character, y can be any string).

Thanks.

### Configuration

_No response_

### Version

0.24.1

### Debug Output

_No response_

### Example Data

_No response_

### Additional Context

_No response_

### References

_No response_

drmason13:
Fixes #725 

Note: the key_value_parser no longer accepts empty keys!

This turned out to be a bit of a rabbit hole, but I tried to make minimal changes to meet fix the immediate issue.

I'm preserving some observed behavior that I've seen (in practice):
* escape sequences are not transformed
* there are no "invalid" escape sequences
* a wide range of keys parse successfully, such as lone quote characters `'` and backslashes `\`

However, I've made one change which may/may not be acceptable:
* empty keys no longer parse successfully

I'd like to think they have limited usefulness and actually indicate an error in the input, but it's up to you. It's simple enough to remove that validation (done using nom's verify() in `parse_key`) but some tests input start parsing weirdly instead of erroring.

Hope that helps, it's been a fun rabbit hole! Kudos for the test suite, was nice and easy to use.

### DIFF
--- a/changelog.d/725.breaking.md
+++ b/changelog.d/725.breaking.md
@@ -1,0 +1,1 @@
+fixed `parse_logfmt` handling of escapes in values that could cause spurious keys to be created. As a result of this fix, the breaking change has been made to no longer allow empty keys in key-value pair formats
--- a/src/stdlib/parse_key_value.rs
+++ b/src/stdlib/parse_key_value.rs
@@ -447,11 +447,14 @@ fn parse_key<'a, E: ParseError<&'a str> + ContextError<&'a str>>(
         })
     } else {
         Box::new(move |input| {
-            alt((
-                parse_delimited('\'', key_value_delimiter),
-                parse_delimited('"', key_value_delimiter),
-                parse_undelimited(key_value_delimiter),
-            ))(input)
+            verify(
+                alt((
+                    parse_delimited('\'', key_value_delimiter),
+                    parse_delimited('"', key_value_delimiter),
+                    parse_undelimited(key_value_delimiter),
+                )),
+                |key: &str| !key.is_empty(),
+            )(input)
         })
     }
 }
@@ -518,6 +521,11 @@ mod test {
         assert_eq!(
             Ok(("", ("key".to_string().into(), "".into()))),
             parse_key_value_::<VerboseError<&str>>("=", " ", Whitespace::Strict, false)("key=")
+        );
+
+        assert!(
+            parse_key_value_::<VerboseError<&str>>("=", " ", Whitespace::Strict, false)("=value")
+                .is_err()
         );
     }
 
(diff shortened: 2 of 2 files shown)


jszwedko:
Thanks @drmason13 ! For:

> empty keys no longer parse successfully

Can you give an example line that no longer parses successfully (or maybe add it as a test that expects an error)?

pront:
Thank you for this PR @drmason13! Regarding the breaking change, we should probably go through the deprecation process unless it's an extreme edge case.

drmason13:
sure, it came up in a roundabout way with this test input:
`level=error field="no quote here ""`
which indeed errors - that last quote is invalid.

but a straight up test case is a good idea, one sec

pront (approved):
Thank you for this PR @drmason13!

Reading through the comments, I think it's worth making this breaking change.

pront:
Please add a changelog fragment. I think it's important to include some details about the breaking case here.

jszwedko:
➕ we should add a changelog fragment for this describing the change (in particular the compatibility break). See https://github.com/vectordotdev/vrl/blob/main/changelog.d/README.md

jszwedko (approved):
Thanks @drmason13 !

PR #746, opened 2024-03-12
fix(parse_nginx_log): allow empty referer in nginx logs (#643)

### LINKED ISSUE #643: parse_nginx_log fails on empty referer
Hello,

it seems pattern definition for nginx common log used by parse_nginx_log function expects non-empty referer: https://github.com/vectordotdev/vrl/blob/2b39353b3236e0aac26314ad47153238c52aa2ff/src/stdlib/log_util.rs#L136C1-L136C106

As it turns out, in practice, referer can be empty, see https://stackoverflow.com/questions/6880659/in-what-cases-will-http-referer-be-empty. E.g. when the enduser
> entered the site URL in browser address bar itself.
    visited the site by a browser-maintained bookmark.
    visited the site as first page in a new window/tab/session, in some browsers.
    clicked a link on a page having restrictive [<meta name="referrer">](http://wiki.whatwg.org/wiki/Meta_referrer) tag.
    clicked a link on a page having restrictive [Referrer-Policy](https://www.w3.org/TR/referrer-policy/) header.
    clicked a link having [rel="noreferrer"](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noreferrer).
    clicked a link in an external application (i.e. not a webbrowser, e.g. Flash).
    switched from a https URL to a http URL.
    has security software installed (antivirus/firewall/etc) which strips the referrer from all requests.
    is behind a proxy which strips the referrer from all requests.
    visited the site programmatically (like, curl) without setting the referrer header (bots!).

This means that any time we get client request with empty referer, vector fails to parse nginx log line. We do get thousands of these issues each day.

Example code for vrl playground (vrl 0.9.1, vector cebe6284).

Working:
https://playground.vrl.dev/?state=eyJwcm9ncmFtIjoic3RydWN0dXJlZCA9IHBhcnNlX25naW54X2xvZyEoLm1lc3NhZ2UsXCJpbmdyZXNzX3Vwc3RyZWFtaW5mb1wiKVxuLiA9IG1lcmdlKC4sIHN0cnVjdHVyZWQpXG4iLCJldmVudCI6eyJtZXNzYWdlIjoiLSAtIC0gWzAzL09jdC8yMDIzOjE0OjIxOjM2ICswMDAwXSBcIlBPU1QgLyBIVFRQLzEuMVwiIDQ5OSAwIFwiLVwiIFwiLVwiIDExMjggMC4wMDMgW3NvbWUuYWRkcmVzcy5jb21dIFstXSBodHRwcyAwIDAuMDA0IDAwNSAxMC41My4xMzQuNDcifSwiaXNfanNvbmwiOmZhbHNlLCJlcnJvciI6bnVsbH0%3D

Broken: 
https://playground.vrl.dev/?state=eyJwcm9ncmFtIjoic3RydWN0dXJlZCA9IHBhcnNlX25naW54X2xvZyEoLm1lc3NhZ2UsXCJpbmdyZXNzX3Vwc3RyZWFtaW5mb1wiKVxuLiA9IG1lcmdlKC4sIHN0cnVjdHVyZWQpXG4iLCJldmVudCI6eyJtZXNzYWdlIjoiLSAtIC0gWzAzL09jdC8yMDIzOjE0OjIxOjM2ICswMDAwXSBcIlBPU1QgLyBIVFRQLzEuMVwiIDQ5OSAwIFwiXCIgXCItXCIgMTEyOCAwLjAwMyBbc29tZS5hZGRyZXNzLmNvbV0gWy1dIGh0dHBzIDAgMC4wMDQgMDA1IDEwLjUzLjEzNC40NyJ9LCJpc19qc29ubCI6ZmFsc2UsImVycm9yIjpudWxsfQ%3D%3D

drmason13:
fixes #643 

The upstreaminfo log format regex was the odd one out on enforcing a non-empty referer. I've switched it to be the same as the others and use a * (0 or more) instead of a + (1 or more) non-" characters.

### DIFF
--- a/changelog.d/643.fix.md
+++ b/changelog.d/643.fix.md
@@ -1,0 +1,1 @@
+`parse_nginx` now accepts empty values for http referer
--- a/src/stdlib/log_util.rs
+++ b/src/stdlib/log_util.rs
@@ -133,7 +133,7 @@ pub(crate) static REGEX_INGRESS_NGINX_UPSTREAMINFO_LOG: Lazy<Regex> = Lazy::new(
         "(?P<request>[^"]*)"\s+                             # Match any non double-quote character
         (?P<status>\d+)\s+                                  # Match numbers
         (?P<body_bytes_size>\d+)\s+                         # Match numbers
-        "(-|(?P<http_referer>[^"]+))"\s+                    # Match `-` or any non double-quote character
+        "(-|(?P<http_referer>[^"]*))"\s+                    # Match `-` or any non double-quote character
         "(-|(?P<http_user_agent>[^"]+))"\s+                 # Match `-` or any non double-quote character
         (?P<request_length>\d+)\s+                          # Match numbers
         (?P<request_time>\d+\.\d+)\s+                       # Match numbers with dot
--- a/src/stdlib/parse_nginx_log.rs
+++ b/src/stdlib/parse_nginx_log.rs
@@ -295,6 +295,23 @@ mod tests {
             tdef: TypeDef::object(kind_combined()).fallible(),
         }
 
+        combined_line_valid_empty_referer {
+            args: func_args![
+                value: r#"0.0.0.0 - - [04/Oct/2022:03:07:27 +0000] "]&\xDF\xBDV\xE7\xBB<\x10;\xA2b}\xDFM\x1D" 400 150 "" "-""#,
+                format: "combined"
+            ],
+            want: Ok(btreemap! {
+                "client" => "0.0.0.0",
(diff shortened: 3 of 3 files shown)


jszwedko (commented):
Makes sense, thanks @drmason13 ! Thanks for the test.

Do you mind adding a changelog fragment? See https://github.com/vectordotdev/vrl/blob/main/changelog.d/README.md

drmason13:
that towncrier thing is nifty, hadn't come across it before. Needs `--config changelog.toml` to run

Would prevent so many merge conflicts at $work

jszwedko (approved):
Thanks @drmason13 !

Issue #720, opened 2024-03-03
Add "dns_lookup" function


johnhtodd:
We have a need to perform lookups on DNS results delivered to us via the DNSTAP source, to determine (as an example) if an error is transient, or is related to a specific geography of resolvers that may be malfunctioning. Also, some DNS requests may yield lookup data in some cases, where a TXT or other result can be used to enrich the logging message as it moves through Vector, or where NS records may be used to enrich data such as "What authoritative nameserver cluster provided these answers?"

We would therefore like to request this as a VRL feature to be added to Vector that allows DNS requests to be emitted from Vector, and the result of those queries to be handed back in some easily parseable way.   The good news is that there already exists a DNSTAP structure, and so this would require grafting an existing Rust-based resolver library to produce results that could be fed to the DNSTAP parser as replies.

There are caveats here, and that is understood. These may be slow processes - the DNS is not easily predicted for response times. Local caches should be used, possibly on the same server as Vector is operating, in order to accelerate results. Timeouts and retries for failures should probably be kept low as a best-practice - "fail fast."  Simply because this may be slow is not sufficient to not perform the lookups - we will probably execute these on a sampled set of events in order to tune an adequate response rate to the stream of total inbound events we need to parse.

This would specifically not allow queries for QTYPEs of AXFR, IXFR, or ANY as they are too complex to easily render and are difficult to imagine in a Vector use case.

Ideally these fields would be used:

 - qname (mandatory)
 - QTYPE (mandatory)
 - namesever destination (optional - if not specified, system resolver will be used)
 - options (optional)
 - class (optional, default IN)

The reply from the resolver code would be handed back as a DNSTAP message structure.

Example:

`.myDnsResult = dns_lookup!("www.example.com", "9.9.9.9", "A", ["timeout=1", "tries=1", "nodnssec", "nsid"])
`

Errors in the DNS lookup should not produce errors within Vector, unless they are invalid command options which cannot be parsed. NXDOMAIN, SERVFAIl, etc. are all legitimate responses from the resolver stack, and should be represented in the DNSTAP message and not raise an error in Vector.  Only if the syntax of the command is incorrect should the lookup fail with an error outcome that is fallible, or if there is a timeout in the response (should this be a fixed number like 5 seconds?) or if there is some problem executing the command like no sockets available or no route to IP address for the recursive resolver.

This option list cribbed from recent Debian "dig" options that seem relevant and which are not redundant:

Options:
```
[...]

PR #764, opened 2024-03-26
feat(stdlib): add `dns_lookup` function

esensar:
Adds `dns_lookup` function, relying on the `domain` crate (`dns_lookup` crate doesn't support any options).

Fixes: #720

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,15 @@
 version = 3
 
 [[package]]
+name = "addr2line"
+version = "0.21.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
+dependencies = [
+ "gimli",
+]
+
+[[package]]
 name = "adler"
 version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -180,6 +189,21 @@ name = "autocfg"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
+
+[[package]]
+name = "backtrace"
+version = "0.3.69"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2089b7e3f35b9dd2d0ed921ead4f6d318c27680d4a5bd167b3ee120edb105837"
+dependencies = [
+ "addr2line",
+ "cc",
+ "cfg-if",
+ "libc",
+ "miniz_oxide",
+ "object",
+ "rustc-demangle",
+]
 
 [[package]]
 name = "base16"
@@ -784,6 +808,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7e962a19be5cfc3f3bf6dd8f61eb50107f356ad6270fbb3ed41476571db78be5"
 
 [[package]]
+name = "deranged"
+version = "0.3.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b42b6fa04a440b495c8b04d0e71b707c585f83cb9cb28cf8cd0d976c315e31b4"
+dependencies = [
+ "powerfmt",
+]
+
+[[package]]
 name = "derive_arbitrary"
 version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -853,8 +886,25 @@ checksum = "e5766087c2235fec47fafa4cfecc81e494ee679d0fd4a59887ea0919bfb0e4fc"
 dependencies = [
  "cfg-if",
  "libc",
- "socket2",
+ "socket2 0.5.3",
  "windows-sys 0.48.0",
+]
+
+[[package]]
+name = "domain"
+version = "0.9.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e853e3f6d4c6e52a4d73a94c1810c66ad71958fbe24934a7119b447f425aed76"
+dependencies = [
+ "bytes",
+ "futures-util",
+ "libc",
+ "octseq",
+ "rand",
+ "serde",
+ "smallvec",
+ "time",
+ "tokio",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


esensar:
I have tried implementing this with `dns_lookup` crate first (which is already included), but it was very limited in its options, so I had used `domain` crate here. `domain` crate pulls in many new dependencies, so I am not sure if that is acceptable to the project. If not, I can look for other options, or go back to `dns_lookup`, for a more limited functionality.

jszwedko (commented):
@johnhtodd do you mind taking a look at this one to make sure the interface matches what you expected?

jszwedko (commented):
`domain` does pull in quite a few new dependencies including Tokio 😅 @esensar if you are open to it, I wouldn't mind evaluating the other options more thoroughly to see if there is one that is a bit lighter.

esensar:
> `domain` does pull in quite a few new dependencies including Tokio 😅 @esensar if you are open to it, I wouldn't mind evaluating the other options more thoroughly to see if there is one that is a bit lighter.

Yeah, definitely. Seeing tokio made me reconsider it :D

esensar:
I think almost all of these are supported by `domain` crate (and probably the `hickory_client` too), but the issue here is that I haven't been able to find a crate for DNS lookup that doesn't depend on `tokio`...

If using `tokio` is fine in the end, then I would actually prefer `hickory_client` instead, since `hickory_proto` is used in `vector` too.

jszwedko:
> I think almost all of these are supported by `domain` crate (and probably the `hickory_client` too), but the issue here is that I haven't been able to find a crate for DNS lookup that doesn't depend on `tokio`...
> 
> If using `tokio` is fine in the end, then I would actually prefer `hickory_client` instead, since `hickory_proto` is used in `vector` too.

Thanks for researching @esensar . I guess it's not _too_ surprising given they would need to issue asynchronous requests. If `hickory_client` could work to satisfy the needs, I agree that seems preferable since we already have a dependency on `hickory_proto` and so it'd be less of a net-new dependency.

esensar:
> > I think almost all of these are supported by `domain` crate (and probably the `hickory_client` too), but the issue here is that I haven't been able to find a crate for DNS lookup that doesn't depend on `tokio`...
> > If using `tokio` is fine in the end, then I would actually prefer `hickory_client` instead, since `hickory_proto` is used in `vector` too.
> 
> Thanks for researching @esensar . I guess it's not _too_ surprising given they would need to issue asynchronous requests. If `hickory_client` could work to satisfy the needs, I agree that seems preferable since we already have a dependency on `hickory_proto` and so it'd be less of a net-new dependency.

I have started moving this to `hickory_client`. It is mostly done, other than adding support for some of the options (it will not be as straightforward as with `domain`).

`hickory_client` does not parse `/etc/resolv.conf` by default, while `domain` did it. Should something like that be supported or should we make `nameserver` required, because `domain` was defaulting to whatever was defined in `/etc/resolv.conf`?

johnhtodd:
> `hickory_client` does not parse `/etc/resolv.conf` by default, while `domain` did it. Should something like that be supported or should we make `nameserver` required, because `domain` was defaulting to whatever was defined in `/etc/resolv.conf`?

I don't think it is a problem to always specify a resolver in the command, so if that is the solution then I'd say that's reasonable. Using /etc/resolv.conf would have been convenient as a default, but that is not a requirement.

jszwedko:
I converted this back to draft since it seems like you are still iterating on it, but feel free to mark it ready when it is ready for review. Thanks for your work on this one!

esensar:
Sorry for once again bringing up the crate concern here, but after spending some time trying to get `hickory_client` to work here, I am not sure if it will be beneficial in the end. To properly support all required options, this requires adding some more dependencies explicitly and generally more complex and harder to maintain function.

On the other hand `domain` crate can't support DoH and DoT, whereas it can be done with `hickory_client` (although it complicates things a bit, it is doable).

Not sure how to progress on this, I am worried about making this function unnecessarily complex and hard to maintain. If DoH and DoT are not critical to be included, I believe going the `domain` route would be safer, even if it resulted in more new dependencies when compared to `hickory_client`.

pront:
>  I believe going the domain route would be safer, even if it resulted in more new dependencies when compared to hickory_client.

I agree, `domain` looks more mature for this use case.

johnhtodd:
> Sorry for once again bringing up the crate concern here, but after spending some time trying to get `hickory_client` to work here, I am not sure if it will be beneficial in the end. To properly support all required options, this requires adding some more dependencies explicitly and generally more complex and harder to maintain function.
> 
> On the other hand `domain` crate can't support DoH and DoT, whereas it can be done with `hickory_client` (although it complicates things a bit, it is doable).
> 
> Not sure how to progress on this, I am worried about making this function unnecessarily complex and hard to maintain. If DoH and DoT are not critical to be included, I believe going the `domain` route would be safer, even if it resulted in more new dependencies when compared to `hickory_client`.

DoT and DoH are not absolutely required, so if `domain` doesn't support those it is not a critical requirement from my perspective.   Simpler = better as long as the core functionality and the mandatory flags/functions are available.

esensar:
Unfortunately, I don't see a way to control dnssec in `domain`, or adding EDNS client subnet. EDNS EDE and NSID should be in OPT record which is returned by this function. Other than that, I think other required options are covered.

johnhtodd:
> Unfortunately, I don't see a way to control dnssec in `domain`, or adding EDNS client subnet. EDNS EDE and NSID should be in OPT record which is returned by this function. Other than that, I think other required options are covered.

DNSSEC is requested by explicitly setting/unsetting the "DO" bit on the request.

NSID has to be requested explicitly on the outbound connection in order to appear in the reply, though EDNS EDE is returned if EDNS is flagged as being supported in the outbound query (which most modern DNS stacks should do, I assume.)

changelog.d/764.feature.md
```diff
@@ -0,0 +1,4 @@
+Added experimental `dns_lookup` function. It should be used with caution, since it involves network
```
pront, line 1: Perhaps we should put it behind a feature flag, at least until we introduce some function result caching mechanism.

esensar: That makes sense, although doesn't `reverse_dns` also include network calls? Should that one be put behind a network flag too, or does that one utilize some kind of caching?

esensar: Sorry, just checking if I understood correctly. Based on discussion from: https://github.com/vectordotdev/vrl/issues/720 I thought this function was supposed to be undocumented (at least until caching is sorted out), similar to `reverse_dns`.

I can add a warning to make it more visible that it involves network calls and should be avoided in most cases if we want to have it documented in the end though.

pront: for the above question cc @jszwedko

jszwedko: 👍 I'm still fine with adding this but not documenting it, for the time being, until we can sort out a caching strategy.

pront (approved):
Code looks good. One thing I noticed is that the tests are not checking all the possible fields. It would be good to strengthen them.

esensar:
> Code looks good. One thing I noticed is that the tests are not checking all the possible fields. It would be good to strengthen them.

That makes sense, I will expand them.

esensar:
> @esensar is this now in final state? If so, maybe @jszwedko wants to take a final look and then we can merge 🚢

Yeah, I think it is. I didn't plan on adding more tests, unless you think more things should be covered.

pront:
> > @esensar is this now in final state? If so, maybe @jszwedko wants to take a final look and then we can merge 🚢
> 
> Yeah, I think it is. I didn't plan on adding more tests, unless you think more things should be covered.

Test coverage is good now, but I see a few test failures. You can run the CI locally with `./scripts/checks.sh`.

esensar:
> > > @esensar is this now in final state? If so, maybe @jszwedko wants to take a final look and then we can merge 🚢
> > 
> > 
> > Yeah, I think it is. I didn't plan on adding more tests, unless you think more things should be covered.
> 
> Test coverage is good now, but I see a few test failures. You can run the CI locally with `./scripts/checks.sh`.

Yeah, I was just about to write about that. I will have to fix that, I don't think it will take long. It shouldn't interfere with review though, if you want to take a final look.

jszwedko (approved):
Nice work @esensar

Issue #755, opened 2024-03-20
parse_json 


guusvw:
Hey dear VRL-folks 👋 
I hope this is not a double posting, I only found issues regards `parse_json` for large ints, I'm having an issues as well for floats.

We're getting a JSON which contains the field `ts` and has in out test case the value `1626175065.5934923`
after passing it to `parse_json` the object has the value of `1626175065.5934925`, the last digit differs by 2.
I know, floats are hard & annoying, but is there some way to avoid/fix it?

```
$ message = "{\"level\":\"info\",\"ts\":1626175065.5934923,\"msg\":\"|200|   32.603µs /healthcheck\",\"statusCode\":200,\"request\":306,\"latency\":0.000032603,\"avgLatency\":0.004298489,\"ipPort\":\"1.1.1.1:12345\",\"method\":\"GET\",\"host\":\"1.1.1.1:8080\",\"path\":\"/healthcheck\",\"referer\":\"\"}"
"{\"level\":\"info\",\"ts\":1626175065.5934923,\"msg\":\"|200|   32.603µs /healthcheck\",\"statusCode\":200,\"request\":306,\"latency\":0.000032603,\"avgLatency\":0.004298489,\"ipPort\":\"1.1.1.1:12345\",\"method\":\"GET\",\"host\":\"1.1.1.1:8080\",\"path\":\"/healthcheck\",\"referer\":\"\"}"

$ obj, err = parse_json(message)

{ "avgLatency": 0.004298489, "host": "1.1.1.1:8080", "ipPort": "1.1.1.1:12345", "latency": 0.000032603, "level": "info", "method": "GET", "msg": "|200|   32.603µs /healthcheck", "path": "/healthcheck", "referer": "", "request": 306, "statusCode": 200, "ts": 1626175065.5934925 }
```

```
❯ bin/vector-0.36.1 --version
vector 0.36.1 (x86_64-apple-darwin 2857180 2024-03-11 14:32:52.417737479)
```

jszwedko:
Hmm, interesting. I know floats have issues with representing certain numbers, but this doesn't appear to be the case for the specific example given: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3d538d9f2ae192da51693bc12fdcfbfa

I believe we just defer to serde_json for deserialization, though, so I'm not sure why it is not preserving the precision.

drmason13:
I jumped on this after seeing that it is in fact solvable through a feature in serde_json.
[...]

PR #767, opened 2024-03-26
fix(stdlib): parse_json improved accuracy float parsing #755

drmason13:
Fixed by activating a feature in serde_json, see https://github.com/serde-rs/json/issues/536

For anyone somehow relying on the innaccurate/faster parsing of floats, it's technically a breaking change.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,7 @@ test_framework = ["compiler", "dep:prettydiff", "dep:serde_json", "dep:ansi_term
 arbitrary = ["dep:quickcheck", "dep:arbitrary"]
 lua = ["dep:mlua"]
 proptest = ["dep:proptest", "dep:proptest-derive"]
+float_roundtrip = ["dep:serde_json", "serde_json/float_roundtrip"]
 
 # Internal testing utilities (used for benches)
 test = ["string_path"]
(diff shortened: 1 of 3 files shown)


pront (changes requested):
Thank you @drmason13. I wonder if a better approach is to add a flag and preserve backwards compatibility. Also the accurate parser is much slower so we might not want to turn it on unconditionally.

pront:
Also cc @jszwedko

jszwedko:
Thanks @drmason13 ! A concern I have is that this apparently comes at 2x overhead cost:

https://github.com/serde-rs/json/blob/b1ebf3888ed728c66c69581ba8d9c4aa7483f486/Cargo.toml#L60-L67

It would be nice to make this an optional feature that could be enabled at runtime but given it is a compiler flag I'm not seeing a way to do that. I'd be curious to see if this shows up in our benchmarks in Vector. I opened a PR to see if it flags any difference:

https://github.com/vectordotdev/vector/pull/20185

drmason13:
The performance implications are a valid concern, interested to see what the benchmarks show.

I think it would get complicated to have this be controlled by a flag. Would you need two versions of the serde_json dependency - one with the feature and one without? I'm not sure how to do this.

jszwedko:
Agreed, the impact on Vector seems fairly negligible, I think likely because I/O dominates the tests rather than the extra CPU spent on encoding/decoding floats.

I would like to see this be an optional feature in the VRL crate, though, that could be enabled by Vector so that other consumers of the VRL crate can decide whether to opt into it. Do you mind refactoring this PR to do that? See https://doc.rust-lang.org/cargo/reference/features.html#dependency-features for what this looks like.

drmason13:
Open to suggestions for the feature name, I've gone with mirroring the serde_json feature name for clarity.

pront (approved):

PR #803, opened 2024-04-18
Add uuid_from_friendly_id function

andrewthad:
Resolves https://github.com/vectordotdev/vrl/issues/799

I've never written any Rust code before, so I recommend reviewing this carefully. It includes a passing test.

### DIFF
--- a/src/stdlib/uuid_from_friendly_id.rs
+++ b/src/stdlib/uuid_from_friendly_id.rs
@@ -1,4 +1,3 @@
-// use std::fmt::format;
 use crate::compiler::prelude::*;
 use crate::stdlib::string_utils::convert_to_string;
 use bytes::Bytes;
@@ -70,7 +69,7 @@ mod tests {
         example_from_docs {
             args: func_args![value: value!("3s87yEvnmkiPBMHsj8bwwc")],
             want: Ok(value!("7f41deed-d5e2-8b5e-7a13-ab4ff93cfad2")),
-            tdef: TypeDef::bytes().infallible(),
+            tdef: TypeDef::bytes().fallible(),
         }
     ];
 }
--- a/src/stdlib/uuid_from_friendly_id.rs
+++ b/src/stdlib/uuid_from_friendly_id.rs
@@ -56,7 +56,7 @@ impl FunctionExpression for UuidFromFriendlyIdFn {
     }
 
     fn type_def(&self, _: &TypeState) -> TypeDef {
-        TypeDef::bytes().infallible()
+        TypeDef::bytes().fallible()
     }
 }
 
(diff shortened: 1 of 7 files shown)


src/stdlib/uuid_from_friendly_id.rs
```diff
@@ -0,0 +1,77 @@
+use crate::compiler::prelude::*;
+use bytes::Bytes;
+use crate::stdlib::string_utils::convert_to_string;
+
+fn uuid_from_friendly_id(value: Value) -> Resolved {
+    let mut buf = [0; 36];
+    let value = convert_to_string(value, true)?;
+    match base62::decode(value) {
+      Err(_) => Err("friendly id decode failure".into()),
```
jszwedko, line 9:
```suggestion
      Err(err) => Err(fmt!("failed to decode friendly id: {}", err).into()),
```

src/stdlib/uuid_from_friendly_id.rs
```diff
@@ -0,0 +1,77 @@
...
+}
+
+impl FunctionExpression for UuidFromFriendlyIdFn {
+    fn resolve(&self, ctx: &mut Context) -> Resolved {
+        let value = self.value.resolve(ctx)?;
+        uuid_from_friendly_id(value)
+    }
+
+    fn type_def(&self, _: &TypeState) -> TypeDef {
+        TypeDef::bytes().infallible()
```
jszwedko, line 59:
```suggestion
        TypeDef::bytes().fallible()
```

This sholuld be marked as fallible because it can fail to decode the input.

jszwedko (commented):
Nice work @andrewthad ! I'm impressed that you jumped into Rust to implement it.

A left some notes, but a couple of other requests:

- Could you add a changelog entry? See https://github.com/vectordotdev/vrl/blob/main/changelog.d/README.md
- Can you add a benchmark? See https://github.com/vectordotdev/vrl/blob/main/benches/stdlib.rs
- After this is merged could you open a follow-up PR on https://github.com/vectordotdev/vector to document the new function? See https://github.com/vectordotdev/vector/pull/20048 for an example

jszwedko (approved):
Great, thanks @andrewthad !

jszwedko:
@andrewthad I pushed a few more tiny fixes. I figured that was faster than pointing them out to you.

andrewthad:
Thanks! Looks good to me.

jszwedko:
@andrewthad merged! Do you think you'd be able to open a follow-up PR on https://github.com/vectordotdev/vector to document the new function? See https://github.com/vectordotdev/vector/pull/20048 for an example

andrewthad:
I've opened a PR for that at https://github.com/vectordotdev/vector/pull/20357

PR #817, opened 2024-05-01
chore: Introduce borrowed path types

bruceg:
For a long time, VRL has had a set of "owned" path types, but no corresponding types outside of `BorrowedSegment`. This makes it harder to name the actual type involved in borrows for implementing other traits.

This PR replaces the tuple types used for borrowed path with `BorrowedTargetPath` and `BorrowedValuePath`, mirroring their `Owned` equivalents. There is one deprecated implementation on the tuple types left in to allow Vector `master` branch to build as-is with these changes.

### DIFF
--- a/src/path/borrowed.rs
+++ b/src/path/borrowed.rs
@@ -100,7 +100,7 @@ impl<'a, 'b> TargetPath<'a> for BorrowedTargetPath<'a, 'b> {
     }
 
     fn value_path(&self) -> Self::ValuePath {
-        self.path.clone()
+        self.path
     }
 }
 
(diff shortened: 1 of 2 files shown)


bruceg:
Noting for posterity that I did try to modify the Vector code to remove the need for `impl ValuePath for (PathPrefix, T)`, but the changes were far from trivial and I wanted to see what others thought of this approach before I put more work into the Vector side.

pront (approved):
LGTM. This is @fuchsnj's favorite code, might worth getting his input before merging.

PR #830, opened 2024-05-07
Fix `std::fmt::Display` for abort expressions

### LINKED ISSUE #829: Abort expression prints incorrectly
Hey all,

I'm currently working on a transpiler from another language into VRL. All other parts of the AST implement Display in a way that is valid syntactically to parse it again i.e. parse <-> display (formatting not withstanding). However the implementation of Display for `abort` incorrectly places a colon in its output.

https://github.com/vectordotdev/vrl/blame/b6fd5ecf3e4d75ccc5a822dcebe8a0fe7d04ef78/src/parser/ast.rs#L1199

The following kind of test succeeds for all other expressions in VRL.

```rs
    #[test]
    fn abort_isomorphism_failure() {
        let expression = r#"abort "error message""#;

        assert_eq!(
            format!("{expression}\n"),
            vrl::parser::parse(expression).unwrap().to_string()
        )
    }
```

I don't know if this kind of isomorphism via `std::fmt::Display` is a guarantee the project is looking to provide, but beyond transpilers it's also very useful for pretty printers since printing from the parser canonicalizes the formatting.

DylanRJohnston:
Fixes #829. I couldn't see any tests for this code, but a very simply one exists as I wrote in the issue. Let me know if you want that included and where it should live.

```rs
    #[test]
    fn abort_display_isomorphism() {
        let expression = r#"abort "error message""#;

        assert_eq!(
            format!("{expression}\n"),
            vrl::parser::parse(expression).unwrap().to_string()
        )
    }
```

### DIFF
--- a/src/parser/ast.rs
+++ b/src/parser/ast.rs
@@ -1196,7 +1196,7 @@ impl fmt::Display for Abort {
             &self
                 .message
                 .as_ref()
-                .map_or_else(|| "abort".to_owned(), |m| format!("abort: {m}")),
+                .map_or_else(|| "abort".to_owned(), |m| format!("abort {m}")),
         )
     }
 }


DylanRJohnston:
Just noticed a few more instances of failing to serialize out re-parsable expressions, namely infailable function calls and queries.

src/parser/ast.rs
```diff
@@ -933,7 +933,7 @@ pub struct Query {
 
 impl fmt::Display for Query {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "{}{}", self.target, self.path)
+        write!(f, "{}.{}", self.target, self.path)
```
DylanRJohnston, line 936: Without this variable queries end up mushed together e.g. `foo.bar` becomes `foobar` however that then causes `.foo.bar` to become `..foo.bar` without the change to `QueryTarget` below.

src/parser/ast.rs
```diff
@@ -958,7 +958,7 @@ impl fmt::Display for QueryTarget {
         match self {
             Internal(v) => v.fmt(f),
             External(prefix) => match prefix {
-                PathPrefix::Event => write!(f, "."),
+                PathPrefix::Event => write!(f, ""),
```
DylanRJohnston, line 961: Should probably just remove this `write!` all together? If I understand correctly, and based on my testing `QueryTarget` always appears with a `.` after it even if `OwnedValuePath` is empty. So this is safe.

pront:
Hi @DylanRJohnston, since this is deep change, it is also worth pointing Vector's VRL dependency to this sha and see if those CI checks are passing.

DylanRJohnston:
Hey @pront apologies I was waiting until the project was complete before coming back to fixing this pull request as I keep uncovering new problems as the project progresses and I didn't want to waste everyone's time with lots of pull requests.

I also made a fairly controversial decision to overload the `fmt::Formatter` width property to carry the indentation level for pretty printing of nested scopes and I wasn't sure you guys would approve of abusing it like that although I can't see any other way that writing a whole separate serialization method to the current use of `fmt::Display`. What do you think?

Issue #678, opened 2024-02-06
Deprecate path coalescing


pront:
https://vector.dev/docs/reference/vrl/expressions/#path-example-nested-path-coalescing

AFAIK this feature is not widely used and not very useful. Deprecating would reduce mental burden for contributors.

PR #836, opened 2024-05-10
chore!: Drop support for coalesce paths

bruceg:
Follow up to #678 

Do we want to merge this already or wait until closer to cutting the next release?


src/path/owned.rs
```diff
@@ -507,52 +430,18 @@ impl Display for OwnedSegment {
...
-                    self.coalesce_i += 1;
-                }
-                coalesce_segment
-            }
-        });
-        if self.coalesce_i == 0 {
-            self.index += 1;
-        }
-        output
+        self.0.next().map(BorrowedSegment::from)
```
pront, line 444: Much better now

pront (commented):
Didn't realize how much code was specific to this feature... looks good at a first glance. I will do a second pass next week. No need to wait IMO.

jszwedko (approved):

pront (approved):

Issue #844, opened 2024-05-18
parse_xml function : text key added when it should not for single nodes


ScriptToddler:
Hello,

## Problem

There appears to be an inconsistency with `text_key` in `parse_xml` function

Whenever a chain of an "even" number of single child is present, the text key seems to be forcibly added even when it should not.
E.g. : 
- `<root><a>test</a></root>` results in `{ "root": { "a": { "text": "test" } } }` instead of expected `{ "root": { "a": "test" } }`
https://playground.vrl.dev/?state=eyJwcm9ncmFtIjoiLnhtbF9wYXJzZWQgPSBwYXJzZV94bWwhKC5tZXNzYWdlKSIsImV2ZW50Ijp7Im1lc3NhZ2UiOiI8cm9vdD48YT50ZXN0PC9hPjwvcm9vdD4ifSwiaXNfanNvbmwiOmZhbHNlLCJlcnJvciI6bnVsbH0%3D
- `<root><a><a1>test</a1></a></root>` results in expected `{ "root": { "a": { "a1": "test" } } }`
https://playground.vrl.dev/?state=eyJwcm9ncmFtIjoiLnhtbF9wYXJzZWQgPSBwYXJzZV94bWwhKC5tZXNzYWdlKSIsImV2ZW50Ijp7Im1lc3NhZ2UiOiI8cm9vdD48YT48YTE%2BdGVzdDwvYTE%2BPC9hPjwvcm9vdD4ifSwiaXNfanNvbmwiOmZhbHNlLCJlcnJvciI6bnVsbH0%3D
- `<root><a><a1><a2>test</a2></a1></a></root>` results in `{ "root": { "a": { "a1": { "a2": { "text": "test" } } } } }` instead of expected `{ "root": { "a": { "a1": { "a2": "test" } } } }`
https://playground.vrl.dev/?state=eyJwcm9ncmFtIjoiLnhtbF9wYXJzZWQgPSBwYXJzZV94bWwhKC5tZXNzYWdlKSIsImV2ZW50Ijp7Im1lc3NhZ2UiOiI8cm9vdD48YT48YTE%2BPGEyPnRlc3Q8L2EyPjwvYTE%2BPC9hPjwvcm9vdD4ifSwiaXNfanNvbmwiOmZhbHNlLCJlcnJvciI6bnVsbH0%3D
- similarly, if you add a sibling to the `<a>` node, the `<a1>` node then is nested with a text key
[...]

PR #849, opened 2024-05-20
Fix proposal for parse_xml in case of unique child:

ScriptToddler:
- Modify single node map insert behaviour -> process_node instead of recurse
- Add tests for unique child situation

Related to https://github.com/vectordotdev/vrl/issues/844


pront:
Hi @ScriptToddler, thank you for this PR! Can you add a test case that demonstrates the bug this PR fixes?

pront:
Also, this will need a changelog fragment, is this a breaking change?

jszwedko:
> Also, this will need a changelog fragment, is this a breaking change?

It is breaking in that it changes the result but I think it is a bug since I'm not sure who would want the old behavior so can just be marked as a fix.

jszwedko:
Small bump on this @ScriptToddler . Just wondering what you think of the above.

ScriptToddler:
> Small bump on this @ScriptToddler . Just wondering what you think of the above.

Hi,

@jszwedko On my part, I consider it should be a fix, as you said.

@pront As for the tests, I did include two new unit tests in the parse_xml.rs file : if_no_sibling and if_no_sibling2, demonstrating that it seems to now work as intended, not adding the unwanted "text" keys. If that does not suffice, I'm not sure what is expected then.

Finally, I must confess I mostly wanted to point out the problem and offer what appeared to me a quick win fix. For my needs, I have found a workaround so I'm not bothered anymore by this behaviour, and I don't intend to spend more time on this. If this PR does not meet your requirements, it can be thrown away, no problem on my part.

Regards,

pront:
Hi @ScriptToddler, thanks again for raising this issue and for submitting a fix. If you don't intend to spend more time on this PR, no worries at all. Leaving it open as a draft for someone else to take over.

pront:
I applied the fix to the latest main and added a changelog.

jszwedko (approved):

PR #859, opened 2024-05-27
enhancement(parse_groks): Add missing Datadog grok filters

vladimir-dd:
Adds some of the missing Datadog grok [filters](https://docs.datadoghq.com/service_management/events/pipelines_and_processors/grok_parser/?tab=filters):

- rubyhash;
- querystring;
- boolean;
- decodeuricomponent;
- xml.


src/datadog/grok/grok_filter.rs
```diff
@@ -195,6 +213,54 @@ pub fn apply_filter(value: &Value, filter: &GrokFilter) -> Result<Value, GrokRun
...
+        GrokFilter::Rubyhash => match value {
+            Value::Bytes(bytes) => parse_ruby_hash(String::from_utf8_lossy(&bytes).as_ref())
+                .map_err(|_e| {
+                    GrokRuntimeError::FailedToApplyFilter(filter.to_string(), value.to_string())
+                }),
+            _ => Err(GrokRuntimeError::FailedToApplyFilter(
+                filter.to_string(),
+                value.to_string(),
+            )),
+        },
```
bruceg, line 225: Could you reduce this repeated pattern to a common function that handles the `match value`, `String` conversion, and error return?

Cargo.toml
```diff
@@ -22,21 +22,22 @@ members = [
 
 
 [features]
-default = ["compiler", "value", "diagnostic", "path", "parser", "stdlib", "datadog", "core"]
+default = ["compiler", "value", "diagnostic", "path", "parser", "parsing", "stdlib", "datadog", "core"]
```
bruceg, line 25: Does the `parsing` feature enable any different features or behavior by itself? I'm just wondering if it even needs to be here, given that the `datadog` feature enables it indirectly.

src/stdlib/parse_query_string.rs
```diff
@@ -1,33 +1,5 @@
...
-                    v => {
-                        *v = Value::Array(vec![v.clone(), value.into()]);
-                    }
-                };
-            })
-            .or_insert_with(|| value.into());
-    }
-    Ok(result.into())
-}
+use crate::parsing::query_string::parse_query_string;
```
bruceg, line 2: Does this imply that the `stdlib` feature needs to depend on `parsing` as well?

bruceg (commented):
Looks pretty reasonable. I have a couple questions about the use of the feature flag and a common code reduction below.

Cargo.toml
vladimir-dd: ah, right, good catch!

Cargo.toml
```diff
@@ -55,6 +56,7 @@ stdlib = [
  "compiler",
  "core",
  "datadog",
+  "parsing",
```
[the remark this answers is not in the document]
vladimir-dd: Updated.

src/stdlib/parse_query_string.rs
vladimir-dd: Right, it does indeed(it is already defined in `Cargo.toml`).

src/datadog/grok/grok_filter.rs
vladimir-dd: good idea, refactored it.

Issue #852, opened 2024-05-22
parse_aws_alb_log failing with new AWS field


Alex-Waring:
AWS have added a new Transaction ID field to their ALB logs, and have not announced this so it's not in their docs yet. However for those who are on the rollout plan, the `parse_aws_alb_log` function is failing:

```
Vector unable to parse alb log: function call error for "parse_aws_alb_log" at (18:45): Log should be fully consumed: " TID_XXXX"
```

The parse_aws_alb_log function should have added to it the ability to conditionally parse this entry, if that's possible, until AWS confirm the rollout is complete.

To workaround this I have added:

```
strip_tid, err = split(.message, " TID_")[0]
if err != null {
    .@error = true
    .@error_msg = "Vector unable to parse alb log: " + err
} else {
    structured, err = parse_aws_alb_log(strip_tid)
    if err != null {
        .@error = true
        .@error_msg = "Vector unable to parse alb log: " + err
    } else {
        . = merge(., structured)
    }
}
```

jszwedko:
Agreed. Do you have an example of one of the new logs @Alex-Waring ?

Alex-Waring:
I have a heavily redacted version yes:

```
https 2024-05-22T03:48:29.047890Z app/lbid 1.1.1.1:44444 - -1 -1 -1 400 - 38 370 "GET http://2.2.2.2:443/ HTTP/1.1" "-" - - - "-" "-" "-" - 2024-05-22T03:48:29.047000Z "-" "-" "-" "-" "-" "-" "-" TID_[a-z0-9]{32}
```

It's identical to the old setup with the TID tacked on the end

Alex-Waring:
Just to confirm, we should be ignoring these appended logs. This is from AWS:

> When new fields are introduced, they are added to the end of the log entry. You should ignore any fields at the end of the log entry that you were not expecting.

nhlushak:
Hey, we have been also affected by this change from AWS. Do you have any plans to patch this behaviour in nearest time?

PR #862, opened 2024-05-29
Add traceability_id field to alb log parse

skbki:
Apparently AWS added new field to ALB log which broke our ingestion pipeline. Only place I found it is AWS Athena guide to parse ALB logs
https://docs.aws.amazon.com/athena/latest/ug/application-load-balancer-logs.html

Closes: https://github.com/vectordotdev/vrl/issues/852

### DIFF
--- a/changelog.d/862.enhancement.md
+++ b/changelog.d/862.enhancement.md
@@ -1,0 +1,1 @@
+Add traceability_id field support to parse_aws_alb_log
--- a/src/stdlib/parse_aws_alb_log.rs
+++ b/src/stdlib/parse_aws_alb_log.rs
@@ -269,8 +269,8 @@ fn take_quoted1(input: &str) -> SResult<String> {
 }
 
 fn take_tid_or_nothing(input: &str) -> SResult<Option<&str>> {
-    if input.starts_with( " TID_") {
-        let (rest, value) = take_anything(&input)?; 
+    if input.starts_with(" TID_") {
+        let (rest, value) = take_anything(input)?;
         Ok((rest, Some(value)))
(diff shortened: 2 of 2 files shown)


misterek:
This is currently breaking all of our ALB log parsing.

However, is it possible to adjust so that new fields are ignored, not breaking everything?  This is apparently being rolled out slowly region to region, so some amount of logs will be broken.

jszwedko (commented):
Thanks for this contribution @skbki !

I think the `traceability_id` field needs to be optional since it appears that AWS seems to still be rolling out this change and so there may be users where this field doesn't exist yet (or if they are processing older ALB logs). Your change here makes the field required.

I agree with https://github.com/vectordotdev/vrl/pull/862#issuecomment-2142803345 that we should also be conservative and just discard unknown fields to future-proof this function, but that could be handled separately from the PR simply adding support for `traceability_id`.

skbki:
Good point. I've managed to make the field optional

jszwedko (approved):
Thanks @skbki . This looks good to me!

jszwedko (commented):
We'll just need a [changelog entry](https://github.com/vectordotdev/vrl/blob/ef07d02e20e021263ddfffb632c22148214d27ec/CHANGELOG.md) too.

pront:
Hi @skbki, thank you for this contribution! A `cargo fmt` should fix the failing CI check.

jszwedko:
Clippy flagged an issue too (`cargo clippy` should show it locally).

skbki:
Corrected issues, should be fine now

PR #870, opened 2024-06-05
fix(parse_groks): support multibyte characters and date matcher w/o zero-padding

vladimir-dd:
- Fixes a case where a string in the grok pattern contains an escaped special character, followed by a non-ascii character.
- Extends `date` matcher to support patterns w/o zero-padding, e.g.: `%{date("M/d/yy HH:mm:ss z"):field}`


pront (approved):
This looks good to me, thanks! Let's add a changelog entry that explains these changes.

PR #872, opened 2024-06-06
fix(datadog grok):  support more fraction of a second formats

vladimir-dd:
Extends `date` [matcher](https://docs.datadoghq.com/service_management/events/pipelines_and_processors/grok_parser/?tab=matchers#parsing-dates):

- support other fraction characters, not only dot. Example: `%{date("M/d/yy HH:mm:ss,SSS z"):field}`
- support any number of fractional digits for patterns. Example: `%{date("M/d/yy HH:mm:ss.SSSS z"):field}`

### DIFF
--- a/src/datadog/grok/matchers/date.rs
+++ b/src/datadog/grok/matchers/date.rs
@@ -10,7 +10,7 @@ use tracing::warn;
 use super::super::parse_grok::Error as GrokRuntimeError;
 
 /// converts Joda time format to strptime format
-pub fn convert_time_format(format: &str) -> std::result::Result<String, String> {
+pub fn convert_time_format(format: &str) -> Result<String, String> {
     let mut time_format = String::new();
     let mut chars = format.chars().peekable();
     while let Some(&c) = chars.peek() {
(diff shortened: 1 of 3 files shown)


src/datadog/grok/matchers/date.rs
```diff
@@ -239,125 +251,119 @@ pub fn apply_date_filter(value: &Value, filter: &DateFilter) -> Result<Value, Gr
                 value.pop(); // drop Z
                 value.push_str("+0000");
             }
-            match &filter.regex_with_tz {
-                Some(re) => {
-                    let tz = re
-                        .captures(&value)
-                        .ok_or_else(|| {
+            if let Some(tz) = filter
```
vladimir-dd, line 254: There are two real changes here in the `apply_date_filter` function:
 - instead of  the removed `regex_with_tz`, which used to be optional, we check the regex for the named <tz> group.
 - we call `replace_sec_fraction_with_dot(filter, &mut value);` to replace fraction character with dot, since `strftime` format supports only dot-leading format `%.f` for a fraction of a second.
 
The rest is formatting.

The rest is reformatting.

src/datadog/grok/matchers/date.rs
```diff
@@ -27,12 +27,8 @@ pub fn convert_time_format(format: &str) -> std::result::Result<String, String>
                 's' => time_format.push_str("%S"),
                 // fraction of second
                 'S' => {
-                    if time_format.ends_with('.') {
-                        time_format.pop(); // drop .
-                        time_format.push_str("%.f");
-                    } else {
-                        time_format.push_str("%f");
```
pront, line 34: Removing this branch doesn't break any existing test, correct?

pront (approved):
Thanks! I left one question and a few nitpicks. Didn't find any bugs.

PR #874, opened 2024-06-06
chore(tests): Use dns.google in dns_lookup tests instead of localhost

jszwedko:
localhost doesn't seem to resolve consistently on MacOS causing test failures locally. Certainly there is some risk relying on dns.google but it seems minimal. We can swap it out with another domain if/as needed. Failures can be seen here: https://github.com/vectordotdev/vrl/pull/875

Also updates the VRL test runner to correctly diff objects.

Easier to review with whitespace changes disabled since I changed the indentation when updating the fixtures.

### DIFF
--- a/src/stdlib/dns_lookup.rs
+++ b/src/stdlib/dns_lookup.rs
@@ -672,7 +672,7 @@ mod tests {
     }
 
     #[test]
-    #[cfg(linux)]
+    #[cfg(target_os = "linux")]
     // MacOS resolver doesn't always handle localhost
     fn test_localhost() {
         let result = execute_dns_lookup(DnsLookupFn {
--- a/src/stdlib/dns_lookup.rs
+++ b/src/stdlib/dns_lookup.rs
@@ -359,13 +359,17 @@ impl Function for DnsLookup {
                     res.answers = filter(res.answers) -> |_, value| {
                         value.rData == "8.8.8.8"
                     }
+                    # remove class since this is also dynamic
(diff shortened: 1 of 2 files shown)


src/stdlib/dns_lookup.rs
```diff
@@ -648,11 +696,10 @@ mod tests {
     }
 
     #[test]
-    fn test_custom_class_and_type() {
+    fn test_custom_class() {
         let result = execute_dns_lookup(DnsLookupFn {
-            value: expr!("localhost"),
-            class: expr!("*"),
```
jszwedko, line 654: I couldn't get it working with `class` set to `*` so I just changed this test to verify the `type`.

jszwedko:
cc/ @esensar in case you have any ideas on a better way to handle this.

pront (approved):
Thank you 👍

src/test/mod.rs
```diff
@@ -230,7 +230,7 @@ fn process_result(
                 print!("{} (expectation)", Colour::Red.bold().paint("FAILED"));
 
                 if !config.no_diff {
-                    let want = serde_json::to_string_pretty(&test.result.clone()).unwrap();
+                    let want = serde_json::to_string_pretty(&want_value).unwrap();
```
pront, line 233: ✨

esensar:
> cc/ @esensar in case you have any ideas on a better way to handle this.

I haven't tested on MacOS, but this is okay, I don't have an idea on how to do it differently.

PR #875, opened 2024-06-06
chore(ci): Also run tests on MacOS

jszwedko:
https://github.com/vectordotdev/vrl/pull/874 is fixing a recent example where the tests failed to run on MacOS. Add this CI check to catch it in the future.

This'll also help reduce the likelihood of VRL bugs on MacOS.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,7 +41,10 @@ jobs:
         run: ./scripts/format_check.sh
 
   test:
-    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest]
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


pront (approved):

.github/workflows/test.yml
```diff
@@ -41,15 +41,23 @@ jobs:
         run: ./scripts/format_check.sh
 
   test:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
```
pront, line 46: For my own understanding, was this the default strategy?

jszwedko: The default strategy is to abort all other matrix runs when one fails. In this case, I think it is useful to just get all of the results.

pront: Gotcha, agreed.

PR #882, opened 2024-06-07
fix(stdlib): properly detect tokio runtime in `dns_lookup`

esensar:
---

### DIFF
--- a/src/stdlib/dns_lookup.rs
+++ b/src/stdlib/dns_lookup.rs
@@ -4,6 +4,7 @@ use crate::compiler::prelude::*;
 mod non_wasm {
     use std::collections::BTreeMap;
     use std::net::ToSocketAddrs;
+    use std::thread;
     use std::time::Duration;
 
     use domain::base::iana::Class;
@@ -12,6 +13,7 @@ mod non_wasm {
     use domain::resolv::stub::conf::{ResolvConf, ResolvOptions, ServerConf, Transport};
     use domain::resolv::stub::Answer;
     use domain::resolv::StubResolver;
+    use tokio::runtime::Handle;
 
     use crate::compiler::prelude::*;
     use crate::value::Value;
@@ -34,9 +36,18 @@ mod non_wasm {
             .map_err(|err| format!("parsing query class failed: {err}"))?;
 
         let conf = build_options(options.try_object()?)?;
-        let answer = StubResolver::run_with_conf(conf, move |stub| async move {
-            stub.query((host, qtype, qclass)).await
-        })
+        let answer = match Handle::try_current() {
+            Ok(_) => thread::spawn(move || {
+                StubResolver::run_with_conf(conf, move |stub| async move {
+                    stub.query((host, qtype, qclass)).await
+                })
+            })
+            .join()
+            .map_err(|_| format!("starting a thread for the lookup failed"))?,
+            Err(_) => StubResolver::run_with_conf(conf, move |stub| async move {
+                stub.query((host, qtype, qclass)).await
+            }),
+        }
         .map_err(|err| format!("query failed: {err}"))?;
 
         Ok(parse_answer(answer)?.into())
--- a/src/stdlib/dns_lookup.rs
+++ b/src/stdlib/dns_lookup.rs
@@ -40,13 +40,17 @@ mod non_wasm {
             let receiver = Arc::new(Mutex::new(receiver));
             Self {
                 thread: Some(thread::spawn(move || loop {
-                    match receiver.lock().unwrap().recv() {
-                        Ok(handle) => {
-                            let result = (handle.job)();
-                            handle.result.as_ref().send(result).unwrap();
-                        }
-                        Err(_) => todo!(),
-                    }
+                    let handle = receiver
+                        .lock()
+                        .expect("Locking job queue failed")
+                        .recv()
+                        .expect("Worker queue closed");
+                    let result = (handle.job)();
+                    handle
+                        .result
+                        .as_ref()
+                        .send(result)
+                        .expect("Sending result back from worker failed");
                 })),
                 queue: Some(sender),
             }
@@ -64,8 +68,16 @@ mod non_wasm {
                 result: Arc::new(sender),
             };
 
-            self.queue.as_ref().unwrap().send(handle).unwrap();
-            return receiver.lock().unwrap().recv().unwrap();
+            self.queue
+                .as_ref()
+                .expect("Expected queue to be present in the worker")
+                .send(handle)
+                .expect("Submitting job to the queue failed");
+            return receiver
+                .lock()
+                .expect("Locking result receiver failed")
+                .recv()
+                .expect("Job result channel closed");
         }
     }
 
--- a/src/stdlib/dns_lookup.rs
+++ b/src/stdlib/dns_lookup.rs
@@ -1,3 +1,8 @@
+//! Function providing DNS lookup functionality
+//! This function involves network calls and should therefore be avoided in most pipelines
+//! Implementation also relies on a single threaded worker for executing calls, blocking on each
+//! call until result is received, which will greatly reduce performance
+//! Use only if absolutely needed
 use crate::compiler::prelude::*;
 
 #[cfg(not(target_arch = "wasm32"))]
@@ -21,6 +26,9 @@ mod non_wasm {
     use crate::compiler::prelude::*;
     use crate::value::Value;
 
+    // Single threaded worker for executing DNS requests
+    // Currently blocks on each request until result is received
+    // It should be avoided unless absolutely needed
     static WORKER: Lazy<Worker> = Lazy::new(Worker::new);
     const CHANNEL_CAPACITY: usize = 0;
 
@@ -33,6 +41,7 @@ mod non_wasm {
     }
 
     impl Worker {
+        // Creates a thread and 2 channels - one for jobs and one for results
         fn new() -> Self {
             let (sender, receiver) =
                 mpsc::sync_channel::<Job<Result<Answer, Error>>>(CHANNEL_CAPACITY);
@@ -56,6 +65,8 @@ mod non_wasm {
             }
         }
 
+        // Sends a job to the worker
+        // Blocks until result is received
         fn execute<F>(&self, f: F) -> Result<Answer, Error>
         where
             F: FnOnce() -> Result<Answer, Error> + Send + 'static,
@@ -78,6 +89,7 @@ mod non_wasm {
         }
     }
 
+    // Custom drop implementation which stops the started thread
     impl Drop for Worker {
         fn drop(&mut self) {
             drop(self.queue.take());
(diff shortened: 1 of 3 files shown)


esensar:
When testing `dns_lookup` in `vector` in a real environment, we got this error:

```
thread 'vector-worker' panicked at
/cargo/registry/src/index.crates.io-6f17d22bba15001f/domain-0.10.0/src/resolv/stub/mod.rs:287:17:
Cannot start a runtime from within a runtime. This happens because a
function (like `block_on`) attempted to block the current thread while
the thread is being used to drive asynchronous tasks.
note: run with `RUST_BACKTRACE=1` environment variable to display a
backtrace
```

Which happens because VRL is called in a tokio managed thread. The solution here was to move the call to another thread, because internally `domain` crate implement blocking calls by still relyting on `async` function implementation.

Let me know if you have a better solution for this, since this might be too much to spawn a thread for each lookup.

pront:
Hi @esensar, thanks for suggesting a fix.

> since this might be too much to spawn a thread for each lookup.
This might be problematic, could we have a pool or a dedicated thread for this?

src/stdlib/dns_lookup.rs
```diff
@@ -34,9 +36,18 @@ mod non_wasm {
             .map_err(|err| format!("parsing query class failed: {err}"))?;
 
         let conf = build_options(options.try_object()?)?;
-        let answer = StubResolver::run_with_conf(conf, move |stub| async move {
-            stub.query((host, qtype, qclass)).await
-        })
+        let answer = match Handle::try_current() {
+            Ok(_) => thread::spawn(move || {
```
jszwedko, line 40: I think if it returns `Ok` you can use something like:

```
futures::executor::block_on(async {
        handle
            .spawn(async {})
})
```

to just execute in the current Tokio runtime instead of spawning a separate thread per https://stackoverflow.com/a/62536772

jszwedko (commented):
I'm realizing this might be the first example of an "async VRL function". Per https://github.com/vectordotdev/vector/pull/20495 we've tried to avoid that until now. I think we can just hack it as you currently have with sync lookups but if we gather more examples like this it might motivate adding `async` VRL function variants (internally, not exposing the async nature to users).

src/stdlib/dns_lookup.rs
esensar: I still have to await on return value of `block_on` which bring us back to the same issue. I think we will have to go with a pool or a dedicated thread until proper async support is added to VRL (if it ever gets added, since it maybe clashes with original goals).

jszwedko: Ah gotcha. Then yeah, a pool seems like a good approach. If/when VRL functions become async we can revisit this. I was going to suggest we thread the through the runtime from Vector into VRL, but I think I'd prefer to just make the functions async than do that.

jszwedko:
@esensar just a heads up that we'll be releasing Vector next week in-case you want to try to fix this before then.

src/stdlib/dns_lookup.rs
```diff
@@ -12,10 +15,69 @@ mod non_wasm {
...
+    }
+
+    impl Drop for Worker {
+        fn drop(&mut self) {
+            drop(self.queue.take());
+            if let Some(thread) = self.thread.take() {
+                thread.join().unwrap();
+            }
+        }
+    }
```
pront, line 79: Hi @esensar, thank you for this patch. I think it's acceptable workaround for an undocumented function. I think it's good to add some comments here and also a module-level doc  `//!` to explain what this function does and the limitations of a single threading solution. Especially when it comes to indefinite blocking.

src/stdlib/dns_lookup.rs
```diff
@@ -12,10 +15,69 @@ mod non_wasm {
...
+            let (sender, receiver) = mpsc::channel::<JobHandle<Result<Answer, Error>>>();
+            let receiver = Arc::new(Mutex::new(receiver));
+            Self {
+                thread: Some(thread::spawn(move || loop {
+                    match receiver.lock().unwrap().recv() {
+                        Ok(handle) => {
+                            let result = (handle.job)();
+                            handle.result.as_ref().send(result).unwrap();
+                        }
+                        Err(_) => todo!(),
```
jszwedko, line 48: What should we do here? Panic?

src/stdlib/dns_lookup.rs
pront: Another reasonable improvement here is to use a bounded channel.

src/stdlib/dns_lookup.rs
```diff
@@ -12,10 +15,69 @@ mod non_wasm {
...
+        job: Job<T>,
+        result: Arc<mpsc::Sender<T>>,
+    }
+
+    struct Worker {
+        thread: Option<thread::JoinHandle<()>>,
+        queue: Option<mpsc::Sender<JobHandle<Result<Answer, Error>>>>,
+    }
+
+    impl Worker {
```
jszwedko, line 37: Did you borrow this implementation from somewhere? It's slightly more complicated than I might have expected.

src/stdlib/dns_lookup.rs
```diff
@@ -12,10 +15,69 @@ mod non_wasm {
...
+        where
+            F: FnOnce() -> Result<Answer, Error> + Send + 'static,
+        {
+            let job = Box::new(f);
+            let (sender, receiver) = mpsc::channel();
+            let receiver = Arc::new(Mutex::new(receiver));
+            let handle = JobHandle {
+                job,
+                result: Arc::new(sender),
+            };
```
jszwedko, line 65: Similarly to my above question, did you borrow this from somewhere? It seems like a lot of "work" to do for each execute function such that I'm wondering what the performance looks like. Maybe we could add a benchmark for it in `benches/stdlib.rs` to see?

jszwedko (commented):
Thanks @esensar . I think we are ok including a single-threaded implementation to unblock use of this function for the release while we think about a better model. I left some inline comments below, but we'd also like to see a warning added as a rustdoc just to make people aware of the potential bottleneck. Do you think you could add that?

src/stdlib/dns_lookup.rs
esensar: It is roughly based on https://doc.rust-lang.org/book/ch20-02-multithreaded.html
I tried to simplify it as much as I could, but I thought I had to have some kind of channel to take in the function and to send back the result.

src/stdlib/dns_lookup.rs
esensar: Alright, that makes sense. I will try to add these as soon as possible, but I will not be available for the next 2-3 days. When is the release planned? I would love to get this ready for that release.

src/stdlib/dns_lookup.rs
jszwedko: Ah, gotcha, yeah I see. It seems like that example just creates one channel that is reused where this implementation creates a channel per call to `execute`.  Is there a reason we need to create one channel per `execute` call?

src/stdlib/dns_lookup.rs
jszwedko: The release is planned for the 17th, which doesn't leave us too much time 😓  We do minor releases every 6 weeks though so it wouldn't have to wait too long if it didn't make it.

src/stdlib/dns_lookup.rs
esensar: Oh, right. My bad, one should be enough.

pront: Right, a single `mpsc::sync_channel(CHANNEL_CAPACITY)` should be enough.
Both the sender and the receiver should be lazily instantiated instances.

esensar: It is all contained in `Worker` instance now, so both of them are lazily insttantiated (since `Worker` is lazily instantiated). There are 2 bounded channels though, one for jobs and one for results. I hope that is alright.

I have made the capacity 0 for testing (meaning it is always blocking), but on the other hand, not sure if it makes sense to make it any bigger, considering there is only 1 thread handling jobs. I might be missing something,

src/stdlib/dns_lookup.rs
esensar: There are already some benchmarks there for the `dns_lookup` function. Did you mean we should add some benchmarks for the `Worker` itself?

jszwedko (approved):
Thanks for the updates @esensar ! I think this looks good to me know.

It seems possible that having a higher channel capacity could improve performance by letting the worker thread process more messages before it switches out, but we'd need to do some benchmarking to prove that out.

jszwedko:
I'd like to get @pront's review again too.

src/stdlib/dns_lookup.rs
```diff
@@ -12,10 +20,86 @@ mod non_wasm {
...
+    use tokio::runtime::Handle;
 
     use crate::compiler::prelude::*;
     use crate::value::Value;
 
+    // Single threaded worker for executing DNS requests
+    // Currently blocks on each request until result is received
+    // It should be avoided unless absolutely needed
+    static WORKER: Lazy<Worker> = Lazy::new(Worker::new);
+    const CHANNEL_CAPACITY: usize = 0;
```
pront, line 33: Let's set a sufficiently large number here?

pront:
> I'd like to get @pront's review again too.

This is significantly improved, thanks @esensar. I had a question about the channel capacity which I believe is important to resolve before we merge this.

src/stdlib/dns_lookup.rs
esensar: I had trouble figuring out the right number for this and then I thought it would make sense to keep it 0, considering that it is just 1 thread, but that is probably wrong.

Do you have suggestion for a number?

pront: What you have works but we could probably go a bit further by allowing some buffering (without worrying about memory explosion), based on the [docs](https://doc.rust-lang.org/std/sync/mpsc/fn.sync_channel.html):

```
This channel has an internal buffer on which messages will be queued. bound specifies the buffer size. When the internal buffer becomes full, future sends will block waiting for the buffer to open up. 
Note that a buffer size of 0 is valid, in which case this becomes “rendezvous channel” where each [send](https://doc.rust-lang.org/std/sync/mpsc/struct.SyncSender.html#method.send) will not return until a [recv](https://doc.rust-lang.org/std/sync/mpsc/struct.Receiver.html#method.recv) is paired with it.
```

jszwedko: Tying it to the number of concurrent executions of the VRL runtime could make sense. For example, the `remap` transform runs one transform per available thread: https://github.com/vectordotdev/vector/blob/b3276b4cc73dee6d3854469562f1b1fcf15a419c/src/topology/builder.rs#L68-L73. To do that "right" though it should probably be configurable on the VRL runtime.

esensar: > Tying it to the number of concurrent executions of the VRL runtime could make sense. For example, the `remap` transform runs one transform per available thread: https://github.com/vectordotdev/vector/blob/b3276b4cc73dee6d3854469562f1b1fcf15a419c/src/topology/builder.rs#L68-L73. To do that "right" though it should probably be configurable on the VRL runtime.

Sorry for just going for the hard coded number, but I thought we could go for that configurable approach when we update this to use multiple threads (or hopefully sometime in the future make it properly async).

pront (approved):
Thank you @esensar!

PR #884, opened 2024-06-10
ci: create a release-plz action

### LINKED ISSUE #880: Install release-plz to simplify release procedure
In order to simplify the release workflow, we should create a github action with [release-plz](https://github.com/MarcoIeni/release-plz).
This way, it will automagically open a PR to release with the adequate version.

jdrouet:
This PR creates a configuration for [release-plz](https://github.com/MarcoIeni/release-plz).
To make this work, we need to follow the first 2 steps from [this doc](https://release-plz.ieni.dev/docs/github/quickstart). This will add a secret token that will allow this new action to publish the new release to crates.io.

Fixes #880

### DIFF
--- a/.github/workflows/release-plz.yml
+++ b/.github/workflows/release-plz.yml
@@ -1,0 +1,26 @@
+name: release-plz
+
+permissions:
+  pull-requests: write
+  contents: write
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  release-plz:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - name: run release-plz
+        uses: MarcoIeni/release-plz-action@v0.5
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
--- a/.release-plz.toml
+++ b/.release-plz.toml
@@ -1,0 +1,11 @@
+[workspace]
+# prevent release-plz to update the changelog. This is handled
+# manually by the developers.
+changelog_update = false
+# prevent the CI to break due to a missing changelog update.
+pr_labels = ["no-changelog"]
+
+[[package]]
+name = "vrl"
+changelog_path = "./CHANGELOG.md"
+changelog_update = false


pront:
Hey @jdrouet, thanks for this. Is this a replacement for just publishing to crates.io? Could we also make it publish a [version tag](https://github.com/vectordotdev/vrl/tree/v0.16.0)?

jdrouet:
> Hey @jdrouet, thanks for this. Is this a replacement for just publishing to crates.io? Could we also make it publish a [version tag](https://github.com/vectordotdev/vrl/tree/v0.16.0)?

It should generate version tags, readmes and publish to crates.io. The complete process.

pront:
> To make this work, we need to follow the first 2 steps from [this doc](https://release-plz.ieni.dev/docs/github/quickstart). This will add a secret token that will allow this new action to publish the new release to crates.io.

Hi @jdrouet, this looks promising. Are you willing to followup with the remaining steps and demonstrate how to create the next VRL release?

jszwedko (commented):
Nice, I agree this looks promising. A couple of questions:

- Would this let us delete some of the release scripts currently in `scripts/`?
- It seems like this publishes on every push to `main`. Do we want that?
- Agreed with @pront that we should document this. I'd suggest doing so in the README

jdrouet:
Sorry to only reply right now.

> Nice, I agree this looks promising. A couple of questions:
> 
> * Would this let us delete some of the release scripts currently in `scripts/`?

I think, we should let `release-plz` do everything, not use custom things, so yes, we should delete those scripts.

> * It seems like this publishes on every push to `main`. Do we want that?

No, it's more magic than that. It will run on every commit on the main branch to generate a new release PR but it will release only when merging that release PR. ([Here is an example](https://github.com/jdrouet/mrml/pull/417))

In that release PR, it will update the changelog, update the version of the package according to the semantic versioning stuff. You can modify the PR if you don't agree with the picked version or if you wanna change the changelog.

> * Agreed with @pront that we should document this. I'd suggest doing so in the README

Could you clarify what you would expect in the readme?
If we want to use this, we should:
- make sure we have a crates.io token set in the github-action configuration
- create the `release-plz` action in our workflow
- check that it properly creates the release PR
- merge the first one to create a release. If it doesn't work, we can still release it manually.

But first, we could also try using the release-plz binary locally to see if it works properly with VRL (never done that locally, I always go rogue on the CI for my project, but I understand we cannot do that here).

jdrouet:
After discussing with @pront and @jszwedko , the way the changelog is generated was brought. In conclusion, we should disable the changelog generation.

jszwedko (approved):
Seems worth trying out to me 👍 I'll let @pront review too.

Issue #137, opened 2021-01-19
Add function memoization


JeanMertz:
There have been several situations in which a VRL function has relatively slow runtime performance, but the operation it performs results in a value that is relatively static.

see: https://github.com/timberio/vector/pull/6141

I propose we add a “function cache” to the VRL runtime.

The idea is this:

- We add a new `state::FunctionCache` object.
- Functions have access to their own piece of this cache, namespaces by the function identifier (to start)
- They can store a `Value` type in the cache to fetch when needed.
- We either implement a TTL in the cache, or a “query counter”. Alternatively we leave cache invalidation up to the function.

There are still some things to work out here, but it seems a valuable property to have in the runtime.

fpytloun:
I am also very interested in this one. Any time soon? 🙂

pront:
> I am also very interested in this one. Any time soon? 🙂

We do not have capacity in the near future. This sounds like a medium size project and will some thought on how to integrate this in the VRL runtime. 

Contributions are always welcome.

PR #900, opened 2024-06-20
Allow expressions as file_desc and message_type args

nb-mouse:
This might be handy to dynamically load a protobuf description files and specifying message type based on payload being transformed.

This implementation allows to pass expressions instead of literal-only (constant) arguments.

Use case:

```toml
[sources.protobuf]
type = "stdin"
decoding.codec = "protobuf"
decoding.protobuf.desc_file = "event.desc"
decoding.protobuf.message_type = "EventLog"

[transforms.transform]
type = "remap"
inputs = ["protobuf"]
source = """
message_type = replace(.event.proto_data.type_url, "type.googleapis.com/", "") ?? ""
file_desc = sha1(message_type) + ".desc"
. = {
  "id": .event.id,
  "type": message_type,
  "name": parse_proto(.event.proto_data.value, file_desc, message_type)
} ?? {}
"""
```

PS. I have a follow-up for recursive handling Any messages when processing source configuration.

### DIFF
--- a/src/stdlib/parse_proto.rs
+++ b/src/stdlib/parse_proto.rs
@@ -76,39 +76,38 @@ impl Function for ParseProto {
 
     fn compile(
         &self,
-        state: &state::TypeState,
+        _state: &state::TypeState,
         _ctx: &mut FunctionCompileContext,
         arguments: ArgumentList,
     ) -> Compiled {
         let value = arguments.required("value");
-        let desc_file = arguments.required_literal("desc_file", state)?;
-        let desc_file_str = desc_file
-            .try_bytes_utf8_lossy()
-            .expect("descriptor file must be a string");
-        let message_type = arguments.required_literal("message_type", state)?;
-        let message_type_str = message_type
-            .try_bytes_utf8_lossy()
-            .expect("message_type must be a string");
-        let os_string: OsString = desc_file_str.into_owned().into();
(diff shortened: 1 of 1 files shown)


jszwedko (commented):
Thanks for this contribution @nb-mouse . The reason these parameters are currently required to be static is so that they can just be loaded once at start-up and reused. This change will require the files to be reloaded on each and every event flowing through VRL and are Vector is likely to incur a large performance overhead because of that. If we implemented https://github.com/vectordotdev/vrl/issues/137 that could enable something like this while maintaining performance. https://github.com/vectordotdev/vrl/blob/main/DESIGN.md discusses this and other VRL design decisions.

Could you share a bit your use case for using a dynamic proto definition? Would it be possible to have multiple `remap` transforms where you split the input by type beforehand? Or is the complete list of types not known apriori?

Alternatively, and just speculating, maybe we could relax the static constraints on functions with some sort of opt-in flag for users know are willing to make the trade-off. This would need some more discussion before implementation though.

jszwedko:
Thanks for the other PR! We'll take a look at that one.

pront:
> Instead of extra parameters we might introduce separate proto_parse function. To simplify the logic and explicitly distinct functionality.

Hello, I came across this while looking at open VRL PRs. I think this idea is good. Also, it would benefit greatly from https://github.com/vectordotdev/vrl/issues/137.

PR #906, opened 2024-06-24
[OPA-1889] Fix some of the remaining grok discrepancies

vladimir-dd:
Fixes some of the found grok discrepancies/bugs, comparing to Datadog [grok parser](https://docs.datadoghq.com/service_management/events/pipelines_and_processors/grok_parser/):

- `date` matcher - assume the current date, year when the time format does not include them.
- `keyvalue` filter - treat extracted fields as paths: `db.operation` means `{ "db": { "operation": "insert" }}`
-   drop `Value:Null` values from the parsed object

### DIFF
--- a/src/datadog/grok/filters/keyvalue.rs
+++ b/src/datadog/grok/filters/keyvalue.rs
@@ -207,7 +207,7 @@ fn parse_line<'a>(
 }
 
 /// Parses the field_delimiter between the key/value pairs, ignoring surrounding spaces
-fn parse_field_delimiter<'a>(field_delimiter: &'a str) -> impl Fn(&'a str) -> SResult<&'a str> {
+fn parse_field_delimiter(field_delimiter: &str) -> impl Fn(&str) -> SResult<&str> + '_ {
     move |input| {
         if field_delimiter == " " {
             space1(input)
@@ -281,7 +281,7 @@ fn parse_quoted<'a>(
 
 /// A delimited value is all the text until our field_delimiter, or the rest of the string if it is the last value in the line,
 #[inline]
-fn parse_delimited<'a>(field_delimiter: &'a str) -> impl Fn(&'a str) -> SResult<&'a str> {
+fn parse_delimited(field_delimiter: &str) -> impl Fn(&str) -> SResult<&str> + '_ {
     move |input| map(alt((take_until(field_delimiter), rest)), |s: &str| s.trim())(input)
 }
 
@@ -356,7 +356,7 @@ fn parse_value<'a>(
     }
 }
 
-fn parse_number<'a>(field_delimiter: &'a str) -> impl Fn(&'a str) -> SResult<Value> {
+fn parse_number(field_delimiter: &str) -> impl Fn(&str) -> SResult<Value> + '_ {
     move |input| {
         map(
             terminated(
--- a/src/datadog/grok/filters/keyvalue.rs
+++ b/src/datadog/grok/filters/keyvalue.rs
@@ -17,7 +17,6 @@ use nom::{
 use once_cell::sync::Lazy;
 use ordered_float::NotNan;
 use regex::Regex;
-use crate::path::parse_value_path;
 
 use super::super::{
     ast::{Function, FunctionArgument},
@@ -144,7 +143,7 @@ pub fn apply_filter(value: &Value, filter: &KeyValueFilter) -> Result<Value, Gro
                     || matches!(&v, Value::Bytes(b) if b.is_empty())
                     || k.trim().is_empty())
                 {
-                    result.insert(&parse_value_path(&k).unwrap_or_else(|_| crate::owned_value_path!(&k)), v);
+                    result.insert(crate::path!(&k), v);
                 }
             });
             Ok(result)
--- a/src/datadog/grok/filters/keyvalue.rs
+++ b/src/datadog/grok/filters/keyvalue.rs
@@ -17,6 +17,7 @@ use nom::{
 use once_cell::sync::Lazy;
 use ordered_float::NotNan;
 use regex::Regex;
+use crate::path::parse_value_path;
 
 use super::super::{
     ast::{Function, FunctionArgument},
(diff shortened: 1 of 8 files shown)


src/datadog/grok/filters/keyvalue.rs
```diff
@@ -207,7 +207,7 @@ fn parse_line<'a>(
 }
 
 /// Parses the field_delimiter between the key/value pairs, ignoring surrounding spaces
-fn parse_field_delimiter<'a>(field_delimiter: &'a str) -> impl Fn(&'a str) -> SResult<&'a str> {
+fn parse_field_delimiter(field_delimiter: &str) -> impl Fn(&str) -> SResult<&str> + '_ {
```
pront, line 210: nice, much easier to read now

src/datadog/grok/matchers/date.rs
```diff
@@ -241,76 +243,122 @@ pub fn time_format_to_regex(format: &str, with_captures: bool) -> Result<RegexRe
...
+
+    if day_is_missing {
+        strp_format = format!("%-d {}", strp_format);
+    }
+    if month_is_missing {
+        strp_format = format!("%-m {}", strp_format);
+    }
+    if year_is_missing {
+        strp_format = format!("%Y {}", strp_format);
+    }
```
pront, line 262: I don't see any bugs here, however it is a complex piece of code. We can improve readability/maintainability by extracting this in a separate function e.g. 

```
pub fn adjust_strp_format(
    original_format: &str, 
    filter_format: &str
) -> String {
    let mut adjusted_format = filter_format.to_string();

    let year_is_missing = !original_format.contains('y') && !original_format.contains('Y');
    let month_is_missing = !original_format.contains('M');
    let day_is_missing = !filter_format.contains('d');

    if day_is_missing {
        adjusted_format = format!("%-d {}", adjusted_format);
    }
    if month_is_missing {
        adjusted_format = format!("%-m {}", adjusted_format);
    }
    if year_is_missing {
        adjusted_format = format!("%Y {}", adjusted_format);
    }

    adjusted_format
}
```

And then we can add a few tests e.g.
```
#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn test_adjust_strp_format() {
        assert_eq!(adjust_strp_format("", strp_format), "%-d %-m %Y ");
        // ...
    }
}
```

src/datadog/grok/matchers/date.rs
```diff
@@ -241,76 +243,122 @@ pub fn time_format_to_regex(format: &str, with_captures: bool) -> Result<RegexRe
...
+
     let timestamp = match value {
         Value::Bytes(bytes) => {
-            let mut value = String::from_utf8_lossy(bytes).into_owned();
+            let original_value = String::from_utf8_lossy(bytes).into_owned();
```
pront, line 266: Same as above.

src/path/mod.rs
```diff
@@ -261,6 +261,14 @@ impl<'a> ValuePath<'a> for &'a str {
     }
 }
 
+impl<'a> ValuePath<'a> for &'a OwnedValuePath {
```
pront, line 264: What was the motivation for moving this here?

vladimir-dd: I had an issue with the visibility of this trait implementation, but actually it should be in the PR for `keyvalue` filter - I'll double-check if I do need this and, if so, move to that PR. Reverting this change here.

src/datadog/grok/matchers/date.rs
vladimir-dd: Good point, updated.

src/datadog/grok/matchers/date.rs
```diff
@@ -241,120 +243,153 @@ pub fn time_format_to_regex(format: &str, with_captures: bool) -> Result<RegexRe
...
-                value.push_str("+0000");
-            }
-            if let Some(tz) = filter
-                .regex
-                .captures(&value)
-                .and_then(|caps| caps.name("tz"))
-            {
-                let tz = tz.as_str();
-                let tz: Tz = tz.parse().map_err(|error| {
+    let original_value = String::from_utf8_lossy(value.as_bytes().ok_or_else(|| {
```
vladimir-dd, line 246: The only real changes here is `adjust_strp_format_and_value()` function below.

src/datadog/grok/matchers/date.rs
```diff
@@ -241,120 +243,141 @@ pub fn time_format_to_regex(format: &str, with_captures: bool) -> Result<RegexRe
...
+            .ok_or_else(|| {
+                GrokRuntimeError::FailedToApplyFilter(
+                    filter.to_string(),
+                    original_value.to_string(),
+                )
+            })?;
+        Ok(Value::from(
+            Utc.from_utc_datetime(&dt.naive_utc()).timestamp_millis(),
+        ))
+    } else {
```
neuronull, line 290: This is a totally non-blocking comment but an observation- there are a ton of nested if/else, if let  branches in this function and that makes it pretty hard to follow the logic when it gets so large. It'd be helpful to have either helper functions or a comment somewhere that describes the behavior.

src/datadog/grok/parse_grok.rs
```diff
@@ -93,16 +95,51 @@ fn apply_grok_rule(source: &str, grok_rule: &GrokRule) -> Result<Value, Error> {
...
+        _ => {}
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use crate::btreemap;
     use crate::value::Value;
+    use chrono::{DateTime, Datelike, Timelike};
+    // use chrono::{DateTime};
```
neuronull, line 142:
```suggestion
```

neuronull (approved):
Overall LGTM though I noticed CI failures. I appreciate the additional unit tests

src/datadog/grok/matchers/date.rs
vladimir-dd: Totally agree - I added a [task](https://datadoghq.atlassian.net/browse/OPA-1935) to address this issue.

PR #910, opened 2024-06-25
fix(stdlib): clear default servers when setting custom servers for `dns_lookup`

esensar:
---


jszwedko (approved):
Good catch.

PR #940, opened 2024-07-15
fix(repl): do not check for unused expressions

pront:
closes: https://github.com/vectordotdev/vector/issues/20840

This check is not useful in REPL mode. This PR disables the check only for REPL mode.


jszwedko:
Verified that this change works by using this branch in Vector.

jszwedko (approved):

PR #973, opened 2024-08-03
feat(stdlib): add casing functions

### LINKED ISSUE #970: include camelCase in VRL
### A note for the community

### Use Cases

a common operation is translating keys between snake_case, camelCase, PascalCase, etc. It would be nice to have this supported directly in VRL along with existing `downcase` and `upcase` functions.

### Attempted Solutions

Some manual VRL script is the only other option

### Proposal

Have not looked into the implementation details but I would propose adding those API functions which do the transformation. I'm happy to attempt a PR if someone can point me to a good place to start.

### References

_No response_

### Version

0.40.0

leshow:
closes #970


src/stdlib/camelcase.rs
```diff
@@ -0,0 +1,134 @@
...
+
+pub(crate) fn into_case(s: String) -> Result<Case, Box<dyn DiagnosticMessage>> {
+    match s.as_ref() {
+        "camelCase" => Ok(Case::Camel),
+        "PascalCase" => Ok(Case::Pascal),
+        "SREAMING_SNAKE" => Ok(Case::ScreamingSnake),
+        "snake_case" => Ok(Case::Snake),
+        "kebab-case" => Ok(Case::Kebab),
+        _ => Err(Box::new(ExpressionError::from(
+            "must match either camelCase PascalCase, SCREAMING_SNAKE snake_case or kebab-case"
```
jszwedko, line 34: Nit you could use `variants()` here like `fmt!("must match one of {}", variants)`

src/stdlib/pascalcase.rs
```diff
@@ -0,0 +1,112 @@
+use crate::{compiler::prelude::*, stdlib::camelcase::into_case};
+
+use convert_case::{Case, Casing};
+
+use crate::stdlib::camelcase::variants;
```
jszwedko, line 5: Could we move these definitions into a shared module that all of the functions depend on rather than have each function depend on `camelcase`? You could move these functions into a `casing` subdirectory of `src/stdlib` that has a `mod.rs`.

jszwedko (commented):
Nice, thanks @leshow ! This is generally looking good to me. I left some comments below, but I'll let @pront review too.

I think a couple of things this is missing are:

- A changelog fragment (see https://github.com/vectordotdev/vrl/blob/main/changelog.d/README.md)
- Benchmarks (see https://github.com/vectordotdev/vrl/blob/main/benches/stdlib.rs)

src/stdlib/pascalcase.rs
leshow: sure

src/stdlib/camelcase.rs
leshow: done

leshow:
OK I've added screamingsnake kebab and refactored it into a module

jszwedko:
> It can really all be done by one function. I don't know if you have a preference for having a `convertcase(value, to, from)` instead of individual functions in your API. convertcase could technically handle upcasing and downcasing as well. 🤷

Yeah, I was thinking about that too. I'd suggest let's start with these new functions since we already have `uppercase` and `lowercase`. We could always add a `convertcase` function later that just calls one of the others.

leshow:
OK. I added the changelog and bench. Not quite sure how to run it though, `cargo bench -- --ignored` didnt work so you might get a pipeline failure.

jszwedko (approved):
Nice, thanks! I'll let @pront give this a look too.

pront:
Hey, thank you @leshow for this PR! 

A couple of things:
* Run `dd-rust-license-tool write` and commit
* Created a PR in Vector to document the new functions

pront:
Note: Another potential design, was to offer a single VRL function for this kind of string manipulations but there's also precedence with the `upcase` and `downcase` functions. Just noting for posterity.

pront (approved):
LGTM

src/stdlib/casing/mod.rs
```diff
@@ -0,0 +1,70 @@
...
+            variants_msg()
+        ))) as Box<dyn DiagnosticMessage>),
+    }
+}
+
+#[inline]
+pub(crate) fn convert_case(value: Value, to_case: Case, from_case: Option<Case>) -> Resolved {
+    match from_case {
+        Some(case) => Ok(value
+            .try_bytes_utf8_lossy()?
```
jszwedko, line 51: Nit: I just noticed that we are propagating an error here but the function is marked as `infallible`. The error _should_ never be returned because it only happens if `value` isn't `Bytes` but we've already validated that via the type signature. For clarity, I think using `expect()` here would avoid confusion.

jszwedko:
Thanks for this contribution @leshow !

PR #982, opened 2024-08-11
feat(stdlib): Add `parse_influxdb` function

### LINKED ISSUE #976: Add `parse_influxdb` function
With the merge of https://github.com/vectordotdev/vector/pull/19637, we thought of using that decoder instead of a custom lua one (which takes a lot of cpu), but it seems that we have malformed line protocol messages and using the `decoding.codec=influxdb` option in sources does allow to route the malformed messages somewhere else, as the data is being dropped.

The alternative we see is to use a remap transform with a `parse_influxdb` function and `drop_on_abort=true`, so we can route the malformed data (`<component_id>.dropped`) to our custom lua decoder which is less strict.

In order to handle this failing line protocol messages, we think that a `parse_influxdb` vrl function is needed. The implementation should be very similar to https://github.com/vectordotdev/vector/blob/210ff0925d391213556f07bf6ce621967f0368ca/lib/codecs/src/decoding/format/influxdb.rs#L97 

Doubt: The source decoder option is `decoding.codec=influxdb` and not `decoding.codec=line_protocol`, shoud we call this function `parse_influxdb` in order to be consistent with the vector option? or should we change the vector config spec to use `decoding.codec=line_protocol`?

jorgehermo9:
Closes #976

### DIFF
--- a/LICENSE-3rdparty.csv
+++ b/LICENSE-3rdparty.csv
@@ -71,6 +71,7 @@ digest,https://github.com/RustCrypto/traits,MIT OR Apache-2.0,RustCrypto Develop
 dirs-next,https://github.com/xdg-rs/dirs,MIT OR Apache-2.0,The @xdg-rs members
 dirs-sys-next,https://github.com/xdg-rs/dirs/tree/master/dirs-sys,MIT OR Apache-2.0,The @xdg-rs members
 dns-lookup,https://github.com/keeperofdakeys/dns-lookup,MIT OR Apache-2.0,Josh Driver <keeperofdakeys@gmail.com>
+doc-comment,https://github.com/GuillaumeGomez/doc-comment,MIT,Guillaume Gomez <guillaume1.gomez@gmail.com>
 domain,https://github.com/nlnetlabs/domain,BSD-3-Clause,NLnet Labs <dns-team@nlnetlabs.nl>
 dyn-clone,https://github.com/dtolnay/dyn-clone,MIT OR Apache-2.0,David Tolnay <dtolnay@gmail.com>
 either,https://github.com/bluss/either,MIT OR Apache-2.0,bluss
@@ -94,6 +95,7 @@ gimli,https://github.com/gimli-rs/gimli,MIT OR Apache-2.0,The gimli Authors
 grok,https://github.com/daschl/grok,Apache-2.0,Michael Nitschinger <michael@nitschinger.at>
 hashbrown,https://github.com/rust-lang/hashbrown,MIT OR Apache-2.0,Amanieu d'Antras <amanieu@gmail.com>
 heck,https://github.com/withoutboats/heck,MIT OR Apache-2.0,The heck Authors
+heck,https://github.com/withoutboats/heck,MIT OR Apache-2.0,Without Boats <woboats@gmail.com>
 hermit-abi,https://github.com/hermitcore/hermit-rs,MIT OR Apache-2.0,Stefan Lankes
 hermit-abi,https://github.com/hermitcore/rusty-hermit,MIT OR Apache-2.0,Stefan Lankes
 hex,https://github.com/KokaKiwi/rust-hex,MIT OR Apache-2.0,KokaKiwi <kokakiwi@kokakiwi.net>
@@ -104,6 +106,7 @@ iana-time-zone,https://github.com/strawlab/iana-time-zone,MIT OR Apache-2.0,"And
 iana-time-zone-haiku,https://github.com/strawlab/iana-time-zone,MIT OR Apache-2.0,René Kijewski <crates.io@k6i.de>
 indexmap,https://github.com/indexmap-rs/indexmap,Apache-2.0 OR MIT,The indexmap Authors
 indoc,https://github.com/dtolnay/indoc,MIT OR Apache-2.0,David Tolnay <dtolnay@gmail.com>
+influxdb-line-protocol,https://github.com/influxdata/influxdb_iox/tree/main/influxdb_line_protocol,MIT OR Apache-2.0,InfluxDB IOx Project Developers
 inout,https://github.com/RustCrypto/utils,MIT OR Apache-2.0,RustCrypto Developers
 io-lifetimes,https://github.com/sunfishcode/io-lifetimes,Apache-2.0 WITH LLVM-exception OR Apache-2.0 OR MIT,Dan Gohman <dev@sunfishcode.online>
 is-terminal,https://github.com/sunfishcode/is-terminal,MIT,"softprops <d.tangren@gmail.com>, Dan Gohman <dev@sunfishcode.online>"
--- a/changelog.d/976.feature.md
+++ b/changelog.d/976.feature.md
@@ -1,0 +1,3 @@
+Added `parse_influxdb` function
+
+authors: jorgehermo9
--- a/changelog.d/976.feature.md
+++ b/changelog.d/976.feature.md
@@ -1,3 +1,3 @@
-Added `parse_influxdb` function
+Added `parse_influxdb` function to parse events encoded using the [InfluxDB line protocol](https://docs.influxdata.com/influxdb/cloud/reference/syntax/line-protocol/).
 
 authors: jorgehermo9
--- a/src/stdlib/parse_influxdb.rs
+++ b/src/stdlib/parse_influxdb.rs
@@ -122,8 +122,76 @@ impl Function for ParseInfluxDB {
     }
 
     fn examples(&self) -> &'static [Example] {
-        // TODO: add examples
-        &[]
+        &[Example {
+            title: "parse influxdb line protocol",
+            source: indoc! {r#"
+                    parse_influxdb("cpu,host=A,region=us-west usage_system=64i,usage_user=10u,temperature=50.5,on=true,sleep=false 1590488773254420000")
+            "#},
+            result: Ok(indoc! {r#"
+                [
+                    {
+                        "name": "cpu_usage_system",
+                        "tags": {
+                            "host": "A",
+                            "region": "us-west"
+                        },
+                        "timestamp": "2020-05-26T14:52:53.254420000Z",
+                        "kind": "absolute",
+                        "gauge": {
+                            "value": 64.0
+                        }
+                    },
+                    {
+                        "name": "cpu_usage_user",
+                        "tags": {
+                            "host": "A",
+                            "region": "us-west"
+                        },
+                        "timestamp": "2020-05-26T14:52:53.254420000Z",
+                        "kind": "absolute",
+                        "gauge": {
+                            "value": 10.0
+                        }
+                    },
+                    {
+                        "name": "cpu_temperature",
+                        "tags": {
+                            "host": "A",
+                            "region": "us-west"
+                        },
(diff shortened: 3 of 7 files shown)


src/stdlib/parse_influxdb.rs
```diff
@@ -0,0 +1,332 @@
+use std::collections::BTreeMap;
+
+use chrono::DateTime;
+use influxdb_line_protocol::{FieldValue, ParsedLine};
+
+use crate::compiler::prelude::*;
+
+fn influxdb_lines_to_metrics(line: ParsedLine) -> Vec<ObjectMap> {
```
jorgehermo9, line 8: I tried to match the Vector's `influxdb` decoder implementation
https://github.com/vectordotdev/vector/blob/master/lib/codecs/src/decoding/format/influxdb.rs

src/stdlib/parse_influxdb.rs
```diff
@@ -0,0 +1,332 @@
...
+                FieldValue::F64(v) => v,
+                FieldValue::Boolean(v) => {
+                    if v {
+                        1.0
+                    } else {
+                        0.0
+                    }
+                }
+                // TODO: return none or return Error?
+                FieldValue::String(_) => return None,
```
jorgehermo9, line 41: In the Vector implementation, when a field contains an string value, it simply ignores it. I don't know if it's okay to silently ignore or it would be better to return an error. 

In my opinion, I would return an error, but then, the Vector's implementation should be changed also so there is no inconsistency between implementations

src/stdlib/parse_influxdb.rs
```diff
@@ -0,0 +1,332 @@
...
+            };
+
+            // influxdb_line_protocol crate seems to not allow NaN float values while parsing
+            // field values and this case should not happen, but just in case, we should
+            // ignore the field.
+            let Ok(field_value) = NotNan::new(field_value) else {
+                // TODO: return none or return error?
+                return None;
```
jorgehermo9, line 49: Same as before, should we silently ignore or return an error?

I don't think this case is even possible, as I couldn't parse a `NaN`  field value with `influxdb_line_protocol` crate.

src/stdlib/parse_influxdb.rs
```diff
@@ -0,0 +1,332 @@
...
+        .collect()
+}
+
+fn parse_influxdb(bytes: Value) -> Resolved {
+    let bytes = bytes.try_bytes()?;
+    let line = String::from_utf8_lossy(&bytes);
+    let parsed_line = influxdb_line_protocol::parse_lines(&line);
+
+    let metrics = parsed_line
+        .collect::<Result<Vec<_>, _>>()?
```
jorgehermo9, line 79: I collect into a vec a lot of times (in the Vector implementation the same), in Lines 70, 79, 84. Is it okay? Or what do you recommend to avoid those?

src/stdlib/parse_influxdb.rs
```diff
@@ -0,0 +1,332 @@
...
+        .flat_map(influxdb_lines_to_metrics)
+        .map(Value::from);
+
+    Ok(Value::Array(metrics.collect()))
+}
+
+impl From<influxdb_line_protocol::Error> for ExpressionError {
+    fn from(error: influxdb_line_protocol::Error) -> Self {
+        Self::Error {
+            message: format!("InfluxDB line protocol parsing error: {error}"),
```
jorgehermo9, line 90: I preprended some context so the errores are not clueless from where they come. Thoughts?

src/stdlib/parse_influxdb.rs
```diff
@@ -0,0 +1,332 @@
...
+    let timestamp = timestamp.map(DateTime::from_timestamp_nanos);
+
+    let tags: Option<ObjectMap> = series.tag_set.as_ref().map(|tags| {
+        tags.iter()
+            .map(|t| (t.0.to_string().into(), t.1.to_string().into()))
+            .collect()
+    });
+
+    field_set
+        .iter()
```
jorgehermo9, line 24:
```suggestion
        .into_iter()
```

src/stdlib/parse_influxdb.rs
jorgehermo9, line 8:
```suggestion
fn influxdb_line_to_metrics(line: ParsedLine) -> Vec<ObjectMap> {
```

src/stdlib/parse_influxdb.rs
```diff
@@ -0,0 +1,332 @@
...
+
+fn parse_influxdb(bytes: Value) -> Resolved {
+    let bytes = bytes.try_bytes()?;
+    let line = String::from_utf8_lossy(&bytes);
+    let parsed_line = influxdb_line_protocol::parse_lines(&line);
+
+    let metrics = parsed_line
+        .collect::<Result<Vec<_>, _>>()?
+        .into_iter()
+        .flat_map(influxdb_lines_to_metrics)
```
jorgehermo9, line 81:
```suggestion
        .flat_map(influxdb_line_to_metrics)
```

LICENSE-3rdparty.csv
```diff
@@ -104,6 +106,7 @@ iana-time-zone,https://github.com/strawlab/iana-time-zone,MIT OR Apache-2.0,"And
 iana-time-zone-haiku,https://github.com/strawlab/iana-time-zone,MIT OR Apache-2.0,René Kijewski <crates.io@k6i.de>
 indexmap,https://github.com/indexmap-rs/indexmap,Apache-2.0 OR MIT,The indexmap Authors
 indoc,https://github.com/dtolnay/indoc,MIT OR Apache-2.0,David Tolnay <dtolnay@gmail.com>
+influxdb-line-protocol,https://github.com/influxdata/influxdb_iox/tree/main/influxdb_line_protocol,MIT OR Apache-2.0,InfluxDB IOx Project Developers
```
jorgehermo9, line 109: generated with `dd-rust-license-tool write`

src/stdlib/parse_influxdb.rs
```diff
@@ -0,0 +1,404 @@
...
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::btreemap;
+
+    test_function![
+        parse_influxdb => ParseInfluxDB;
+
+        influxdb_valid {
+            args: func_args![ value: format!("cpu,host=A,region=us-west usage_system=64i,usage_user=10u,temperature=50.5,on=true,sleep=false 1590488773254420000") ],
```
jorgehermo9, line 252: tried to include all possible value cases in this test, as it will be useful too as an example & benchmark.

jorgehermo9:
Opened https://github.com/vectordotdev/vector/pull/21105 to address documentation

changelog.d/976.feature.md
```diff
@@ -0,0 +1,3 @@
+Added `parse_influxdb` function
```
jszwedko, line 1:
```suggestion
Added `parse_influxdb` function to parse events encoded using the [InfluxDB line protocol](https://docs.influxdata.com/influxdb/cloud/reference/syntax/line-protocol/).
```

src/stdlib/parse_influxdb.rs
jszwedko: Hmm, yeah, I also lean towards error since we don't have a metric type in Vector to represent metrics that have a string value. We should update the codec implementation too to be consistent.

src/stdlib/parse_influxdb.rs
jszwedko: Mmm, good question. Looking at the parser I agree that there doesn't seem to be a case where it would return NaN. However, this could change in the future. I lean towards returning an error rather than silently ignoring.

src/stdlib/parse_influxdb.rs
jszwedko: 👍 this seems reasonable to me.

src/stdlib/parse_influxdb.rs
jszwedko: I think you could consider https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.try_fold here to collect until until you hit the first error, but I like that what you currently have returns an error early before trying to do mapping of the parsed lines. I think it's ok to leave as-is. We can optimize it after this is merged if needed.

jszwedko (commented):
Thanks for this @jorgehermo9 !

The implementation seems fairly straight-forward. I think, similar to the codec being the first to only emit metrics, that this is the first function that returns a "metric". As-written, `parse_influxdb` will return a "metric" in the form of a log event. If users want to convert this to an actual metric they'll need to pass this through a `log_to_metric` transform with `all_metrics` set to `true`. I think this is ok as an intermediate step since I think refactoring VRL to have stdlib functions that return metrics is a much larger UX change, but I think we should clearly document this in your [documentation PR](https://github.com/vectordotdev/vector/pull/21105); that users should pass the output through a `log_to_metric` transform to convert the metric represented as a log into an actual metric. Does that make sense?

Thanks for adding this!

jorgehermo9:
Updated the Vector's documentation in https://github.com/vectordotdev/vector/pull/21105/commits/fedbeb28313e1e20ea0e6d356b17e45c0d5cbf31 
Feel free to change wording if the explanation it's not clear
@jszwedko

src/stdlib/parse_influxdb.rs
jorgehermo9: Opened https://github.com/vectordotdev/vector/issues/21133 to track this in the codec implementation

src/stdlib/parse_influxdb.rs
jorgehermo9: Agree!

src/stdlib/parse_influxdb.rs
```diff
@@ -0,0 +1,431 @@
...
+            });
+            metric.insert("gauge".into(), gauge_object);
+
+            Ok(metric)
+        })
+        .collect()
+}
+
+#[derive(Debug, Clone, thiserror::Error)]
+enum Error {
```
jorgehermo9, line 77: Created this error enum for convenience instead of working with raw strings. It is only used in Lines 42 and 50

src/stdlib/parse_influxdb.rs
```diff
@@ -0,0 +1,431 @@
...
+    #[error("field set values of type string are not supported")]
+    StringFieldSetValuesNotSupported,
+    #[error("NaN field set values are not supported")]
+    NaNFieldSetValuesNotSupported,
+}
+
+impl From<Error> for ExpressionError {
+    fn from(error: Error) -> Self {
+        Self::Error {
+            message: format!("Error while converting InfluxDB line protocol metric to Vector's metric model: {error}"),
```
jorgehermo9, line 87: Preprended a different prefix than the From implementation for the `influxdb_line_protocol::Error`. Is this too verbose?

jszwedko: I like it 👍

jszwedko (approved):
Awesome, thanks @jorgehermo9 !

jorgehermo9:
tests failed due to a network issue 😮 @jszwedko

jszwedko:
> tests failed due to a network issue 😮 @jszwedko

Huh, weird, I retried it 🙏

PR #993, opened 2024-08-18
feat(stdlib): add unflatten vrl function

### LINKED ISSUE #81: Add `unflatten` function
### Community Note

### Current Vector Version

```
vector 0.19.1 (x86_64-apple-darwin 3cf70cf 2022-01-25)
```

### Use-cases

We have a `flatten` function that flattens objects by joining together nested keys with `.`. It'd be useful to have a way to reverse this. This could also be useful for sources of data that use `.`s in keys.

See https://github.com/vectordotdev/vector/discussions/11062

### Attempted Solutions

This `lua` transform works:

```toml
[transforms.nest]
type = "lua"
inputs = ["stdin"]
version = "2"
source = '''
function build_array(...)
  local arr = {}
  for v in ... do
    arr[#arr + 1] = v
  end
  return arr
end

function process(event, emit)
  for key, value in pairs(event.log) do
    event.log[key] = value

    parts = build_array(string.gmatch(key, "[^\\.]+"))

    -- key has a dot in it, break it apart to create nested table
    if #parts > 1 then
      event.log[key] = nil

      to_update = event.log
      for i = 1, #parts - 1 do
        if to_update[parts[i]] == nil then
          to_update[parts[i]] = {}
        end
        to_update = to_update[parts[i]]
      end
      to_update[parts[#parts]] = value
    end
  end
  emit(event)
end
'''
hooks.process = "process"
```

### Proposal

We add a `unflatten` function to VRL that does the opposite of `flatten`.

jorgehermo9:
Closes #81 

Had no time to polish this PR, but I think it is good enough to be reviewed.I will do the things I've remaining asap, but pushing this if happens that someone can manage to review it in the meanwhile (no hurry)

Things I've left but couldn't do yet because had no time
- changelog
- better examples
- documentation of this function in a Vector repo PR
- Add tests so we can check that `unflatten` is the inverse of `flatten` in the tests under `lib/tests/tests/functions`

If I miss anything, tell me please

I will address this as soon as I can, but I will be off these days.


Cargo.toml
```diff
@@ -142,7 +142,7 @@ iana-time-zone = { version = "0.1", optional = true }
 idna = { version = "0.5", optional = true }
 indexmap = { version = "~2.4.0", default-features = false, features = ["std"], optional = true}
 indoc = {version = "2", optional = true }
-itertools = { version = "0.13", default-features = false, optional = true }
+itertools = { version = "0.13", default-features = false, features=["use_std"], optional = true }
```
jorgehermo9, line 145: Need that feature to use `into_grouping_by`

src/stdlib/unflatten.rs
```diff
@@ -0,0 +1,279 @@
...
+
+fn do_unflatten(value: Value, separator: &str) -> Value {
+    match value {
+        Value::Object(map) => do_unflatten_entries(map, separator).into(),
+        // Note that objects inside arrays are not unflattened
+        _ => value,
+    }
+}
+
+// this should return the key to insert?
```
jorgehermo9, line 21:
```suggestion
```

jszwedko (commented):
You are on a roll :)

This looks like it is heading in the right direction to me. I appreciate the number of tests that were added.

If you are feeling inspired, this could be a good use-case for property testing where random inputs are fed into `flatten` through `unflatten` and then asserted to be the same. This would be the first time we have these sorts of tests in the stdlib, though, so might not be straight-forward; definitely not a requirement for merging this. There are some prop tests declared for the compiler here: https://github.com/vectordotdev/vrl/tree/main/lib/proptests

lib/tests/tests/functions/flatten/simple.vrl
```diff
@@ -0,0 +1,4 @@
+# object: { "a": { "b": { "c": { "d": 1, "e": 2 }, "f": 3 }, "g": 4 } }
+# result: { "a.b.c.d": 1, "a.b.c.e": 2, "a.b.f": 3, "a.g": 4 }
+
+flatten(.)
```
jorgehermo9, line 4: Maybe this test is redudant with the ones from the examples
https://github.com/vectordotdev/vrl/blob/d3a7dd7fda91f6d200ec3d09b70bc20c5af4867f/src/stdlib/unflatten.rs#L96

What do you think? Should I only leave the `unflatten(flatten(.))` one?

jorgehermo9:
I added a new `recursive` parameter to control wether the values should be unflattened recursively too. I hope the examples explains how it behaves. What do you think about this?

lib/tests/tests/functions/flatten/from_unflatten.vrl
```diff
@@ -0,0 +1,4 @@
+# object: { "a.b.c.d": 1, "a.b.c.e": 2, "a.b.f": 3, "a.g": 4 }
+# result: { "a.b.c.d": 1, "a.b.c.e": 2, "a.b.f": 3, "a.g": 4 }
+
+flatten(unflatten(.))
```
jszwedko, line 4: 👍 I like this test.

lib/tests/tests/functions/flatten/simple.vrl
jszwedko: I think it is still a useful test to assert the unflattened representation. One could imagine a bug that causes round tripping through `flatten` to pass while the output of the `unflatten` call is incorrect.

src/stdlib/unflatten.rs
```diff
@@ -0,0 +1,329 @@
...
+            };
+            (head, rest, value)
+        })
+        .into_group_map_by(|(head, _, _)| head.clone());
+
+    grouped
+        .into_iter()
+        .map(|(key, mut values)| {
+            if values.len() == 1 {
+                match values.pop().unwrap() {
```
jszwedko, line 41:
```suggestion
            if values.len() == 1 {
                match values.pop().expect("exactly one element") {
```

We typically `expect` over `unwrap` to provide more context about why the panic should never be hit.

I think one potential way to refactor this that might avoid this unwrap/expect and avoid the early returns is something like:

```rust
match values {
  [value] => { ... what you currently have for handling if there is only one value ... }
  values => { ... handle multiple values ... }
}
```

src/stdlib/unflatten.rs
```diff
@@ -0,0 +1,329 @@
...
+        _ctx: &mut FunctionCompileContext,
+        arguments: ArgumentList,
+    ) -> Compiled {
+        let value = arguments.required("value");
+        let separator = arguments
+            .optional("separator")
+            .unwrap_or_else(|| expr!(DEFAULT_SEPARATOR));
+        let recursive = arguments
+            .optional("recursive")
+            .unwrap_or_else(|| expr!(false));
```
jszwedko, line 150: Should the default be `true` to match `flatten`?

jszwedko (commented):
Nice, thanks @jorgehermo9 ! I left a few more comments, but this is looking good to me. No worries about the prop tests. I appreciate you giving those a shot, but definitely not required for this.

src/stdlib/unflatten.rs
jorgehermo9: I tried to do that, but the problem is `values`  is a `Vec` and therefore, we should convert it to a slice in order to pattern match against it, using `values[..]`, but then, at the first branch we would have the first element borrowed and we would need it owned. The same with the second branch, we will have a borrowed slice of values, but not the owned values and we need owned values for the `do_unflatten_entries` function.

Does this makes sense?

jorgehermo9:
Hi @jszwedko 

I addressed the comments. I additionally add a more efficient way of constructing objects with "unique paths", for example:
```json
{
  "a.b.c.d.e.f": 1,
  "g.h.i.j.k" 2,
}
```
Since the two entries don't share any prefix, they can be constructed without the recursive calls to `do_unflatten_entries` and then we can construct the objects iteratively with the `do_unflatten_entry`. 

This may be a premature optimization, but seems logical to me to not construct a HashMap for grouping keys recursively if there will be just one key group at each recursive call (see https://github.com/vectordotdev/vrl/pull/993/commits/6a0758ddb0778292e21785264a9a88f2ba667a62)

src/stdlib/unflatten.rs
jszwedko: Mmm, yeah, that makes sense. I wish there was a better way to model this that doesn't require the `expect` but I'm not seeing it.

jszwedko (approved):
Great! Thanks for this contribution @jorgehermo9 !

jorgehermo9:
Sorry, clippy action failed in CI. I though that `script/checks.sh` would run `cargo clippy` but I see that it doesn't.

Could you approve again? @jszwedko

PR #1000, opened 2024-08-22
chore(datadog search): Rewrite parse function in terms of `FromStr`

bruceg:
The `FromStr` interface is a standard in Rust for parsing strings into some
other type, and can then be used through the `str::parse<T>` which is
implemented for all `T` that implement `FromStr`. Most of the lines changed in
this PR come from rewriting tests to use the different interface.

### DIFF
--- a/src/datadog/search/parser.rs
+++ b/src/datadog/search/parser.rs
@@ -1,3 +1,5 @@
+use std::str::FromStr;
+
 use pest::Parser;
 
 use super::{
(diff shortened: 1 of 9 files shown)


src/datadog/search/parser.rs
```diff
@@ -8,35 +10,44 @@ use super::{
...
 }
 
 #[cfg(test)]
 mod tests {
     use super::super::node::{BooleanType, Comparison, ComparisonValue, QueryNode};
     use super::*;
 
+    fn parse(s: &str) -> QueryNode {
+        s.parse()
+            .unwrap_or_else(|_| panic!("Unable to parse {s:?}."))
```
jorgehermo9, line 38: I think this is more idiomatic?
```suggestion
            .expect("Unable to parse {s:?}.")
```

neuronull (approved):

PR #1003, opened 2024-08-23
fix(datadog search): remove Datadog logs intake event structure assumption

bruceg:


neuronull (approved):
one small nit in a code comment otherwise looks LGTM

PR #1007, opened 2024-08-26
chore(datadog search): Allow use of `QueryNode` in configurations

bruceg:
This adds support for serde serialization and deserialization for queries via their string representation, allowing them to be used directly in configurations without storing them in an intermediate `String`.


jszwedko (commented):
Is there anything that would make sense to test here?

neuronull (approved):

PR #1009, opened 2024-08-27
chore(deps): Relax constraint on indexmap

jszwedko:
I can't find a reason this needs to be restricted and it is prohibiting upgrades of indexmap in Vector.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -141,7 +141,7 @@ hex = { version = "0.4", optional = true }
 hmac = { version = "0.12", optional = true }
 iana-time-zone = { version = "0.1", optional = true }
 idna = { version = "0.5", optional = true }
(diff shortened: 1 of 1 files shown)


bruceg:
It was pinned in Vector: https://github.com/vectordotdev/vector/pull/9089
But not unpinned when the comment was removed: https://github.com/vectordotdev/vector/pull/9512

bruceg (approved):

jszwedko:
> It was pinned in Vector: [vectordotdev/vector#9089](https://github.com/vectordotdev/vector/pull/9089) But not unpinned when the comment was removed: [vectordotdev/vector#9512](https://github.com/vectordotdev/vector/pull/9512)

Aha, yes, this is coming back to me now. Thanks for resurfacing that!

PR #1010, opened 2024-08-30
feat(stdlib): Add `parse_auditd` vrl function

### LINKED ISSUE #66: New parser or source for Linux audit logs
### A note for the community

### Use Cases

The Linux kernel can provide a stream of audit logs, which allows system administrators to observe high-level security-related events without needing to trust every process to faithfully log what it is doing. This is accessible through a kernel API, but most Linux operating systems include a userspace service (auditd) to transfer these logs to a human-readable file on disk.

In an environment with hundreds of servers where system security is a major concern, I would like to build tooling to aggregate these audit logs at scale and deliver notifications when unusual activity is detected. Additionally, the audit log records when any operation is denied due to the active SELinux policy, so being able to aggregate these denials across many hosts would be useful when developing and tuning the SELinux policy.

### Attempted Solutions

Obviously you can ingest audit.log using the `file` source, but only as plain text. The data is structured and it would be useful to be able to access the individual fields in Vector for analysis/routing purposes.

At first glance it might look like you could parse it with `parse_key_value`, but it is subtly more complex. For example, given this log line:

```
type=USER_LOGOUT msg=audit(1651746068.966:32475): pid=122482 uid=0 auid=0 ses=146 subj=system_u:system_r:sshd_t:s0-s0:c0.c1023 msg='op=login id=0 exe="/usr/sbin/sshd" hostname=? addr=1.2.3.4 terminal=ssh res=success' UID="root" AUID="root" ID="root"
```

This looks like a really strange format, but makes slightly more sense when you realise that this part:

```
audit(1651746068.966:32475): pid=122482 uid=0 auid=0 ses=146 subj=system_u:system_r:sshd_t:s0-s0:c0.c1023 msg='op=login id=0 exe="/usr/sbin/sshd" hostname=? addr=1.2.3.4 terminal=ssh res=success'
```

is the raw text emitted by the kernel, which auditd substitutes (unquoted) into a template like this:

```
type=USER_LOGOUT msg=<raw line> UID="root" AUID="root" ID="root"
```

If you try to pass this line to `parse_key_value`, it sees two pairs with the same key: `msg=audit(1651746068.966:32475):` and `msg='op=login`. Only the latter is kept; the former (which contains the timestamp!) is lost. This could be worked around by splitting the line somewhere in between e.g. immediately after the `msg=audit(...):` token.

Also, we end up with `msg='op=login` and `res=success'` but `op=login` and `res=success` would be more useful. This is due to `parse_key_value` only recognising double-quoted values, not single-quoted ones. Again, this can be worked around with string manipulation functions.

Currently we're using individual regexes to extract just the fields that we're interested in. It would be nice to have a general solution for parsing these logs.

#### A hacky solution

Extend `parse_key_value` to understand single-quoted values as well as double-quoted ones.
Together with some string splitting, regexing, and timestamp parsing, these log lines can be fully parsed.
A "recipe" for this could be included in the examples section of the docs.

#### A good solution

The above solution requires a lot of VRL code, which everyone who wants to parse this log would have to duplicate. There is already precedent for having many specialised VRL functions for parsing different formats, so we could introduce a new one for parsing the auditd log format.

Example VRL (same input string as discussed above):

```
parse_auditd(s'type=USER_LOGOUT msg=audit(1651746068.966:32475): pid=122482 uid=0 auid=0 ses=146 subj=system_u:system_r:sshd_t:s0-s0:c0.c1023 msg=\'op=login id=0 exe="/usr/sbin/sshd" hostname=? addr=1.2.3.4 terminal=ssh res=success\' UID="root" AUID="root" ID="root"')
```

Return value:

```json
{
  "type": "USER_LOGOUT",
  "timestamp": "2022-05-05T10:21:08.966Z",
  "counter": 32475,
  "pid": 122482,
  "uid": 0,
  "auid": 0,
  "ses": 146,
  "subj": "system_u:system_r:sshd_t:s0-s0:c0.c1023",
  "msg": {
    "op": "login",
    "id": 0,
    "exe": "/usr/bin/sshd",
    "hostname": "?",
    "addr": "1.2.3.4",
    "terminal": "ssh",
    "res": "success"
  },
  "UID": "root",
  "AUID": "root",
  "ID": "root"
}
```

Alternatively the contents of the `msg` object could be flattened into the top-level object. I don't have a preference on this.

#### A better solution

The above solution will break if auditd ever changes its logging format. Since the file it writes to is intended to be human-readable, this is entirely possible.

A more robust (but more complex) solution would be to implement a new source which skips auditd's serialisation and uses the kernel API (via `libaudit.so`) instead.

### References

This may be a duplicate of an existing issue - there's some 200 issues mentioning the word "audit" and I haven't checked all of them.

### Version

vector 0.21.1 (x86_64-unknown-linux-gnu 18787c0 2022-04-22)

jorgehermo9:
Closes #66

I have a few things left (that I'm aware of):

- [ ] Changelog
- [ ] Documentation PR in Vector repo
- [ ] Add this function to benchmarks
- [ ] `linux-auditd-parse` crate should have a license compatible with Vector (see https://github.com/hillu/linux-audit-parser-rs/issues/1)
- [ ] Better examples 
- [ ] Contribute to `linux-auditd-parse` for a few things that are left as TODOs

I added a few TODOs in the code for some things I had doubts. I'll appreciate feedback on those. 

Currently, logs are parsed into something similar to 
```json
{
    "body": {
        "auid": 1000,
        "format": "enriched",
        "kernel": "6.10.4-arch2-1",
        "op": "start",
        "pid": 1240242,
        "res": "success",
        "ses": 2,
        "uid": 0,
        "ver": "4.0.2"
    },
    "enrichment": {
        "AUID": "vrl",
        "UID": "root"
    },
    "sequence": 6439,
    "timestamp": "2024-08-23T14:27:54.618Z",
    "type": "DAEMON_START"
}
```
Which is a bit more structured from what has been proposed in #66, but I'm open to any suggestion about this format


src/stdlib/parse_auditd.rs
```diff
@@ -0,0 +1,432 @@
...
+
+use crate::{btreemap, compiler::prelude::*};
+use bytes::BytesMut;
+use chrono::DateTime;
+use linux_audit_parser::{Number, Value as AuditValue};
+
+fn parse_auditd(bytes: Value) -> Resolved {
+    let bytes = bytes.try_bytes()?;
+    // check if bytes ends with newline, otherwise append it
+    // TODO: make the parser accept bytes without newline in the linux_audit_parser crate (to-be-contributed)
```
jszwedko, line 11: Agreed, it seems like the parser should work without trailing newlines.

src/stdlib/parse_auditd.rs
```diff
@@ -0,0 +1,432 @@
...
+
+    let timestamp_millis = i64::try_from(parsed.id.timestamp)
+        .map_err(|_| Error::TimestampOverflow(parsed.id.timestamp))?;
+    // Should we use UTC as the timezone? The logs are generated with the system's
+    // timezone... What is the correct behavior? Maybe the system where vector is running
+    // have a different timezone than the system that generated the logs... so it is
+    // not correct to assume that the current system's timezone is the correct one
+    // (with TimeZone::timestamp_millis_opt)
+    // TODO: we should document that the timestamp is parsed into UTC timezone.
+    // TODO: Maybe we should accept in a parameter a custom timezone to parse the timestamp?
```
jszwedko, line 31: I think we should rely on the `timezone` that is injected into the VRL runtime. See an example in the `parse_syslog` function: https://github.com/vectordotdev/vrl/blob/c6f22a113253dbbff62abea646955ac62e819913/src/stdlib/parse_syslog.rs#L6-L19

This timezone is configurable in Vector (e.g. on the `remap` transform: https://vector.dev/docs/reference/configuration/transforms/remap/#timezone).

src/stdlib/parse_auditd.rs
```diff
@@ -0,0 +1,432 @@
...
+        log.insert("node".into(), Value::from(Bytes::from(node)));
+    }
+
+    let message_type = parsed.ty.to_string();
+    log.insert("type".into(), Value::from(message_type));
+
+    let (enrichment, body): (ObjectMap, ObjectMap) = parsed
+        .body
+        .into_iter()
+        // TODO: improve this auditd crate with a IntoIter implementation for Body and not only
```
jszwedko, line 50: Agreed 👍

src/stdlib/parse_auditd.rs
```diff
@@ -0,0 +1,432 @@
...
+        Ok(result)
+    }
+}
+
+impl From<Number> for Value {
+    fn from(number: Number) -> Self {
+        match number {
+            Number::Dec(decimal) => Value::from(decimal),
+            // TODO: should we store hexadecimals as its integer value or as an hexadecimal string?
+            // TODO: Uppercase hexa or lowercase hex format?
```
jszwedko, line 119: Good question. Do you happen to know how users would typically want to consume these values or what fields typically contain them? I think the answer to this question likely depends on what sorts of operations users need to do on these values (if any).

jszwedko (commented):
Awesome! I'm happy to see this work started as it is an oft-requested feature. I left a few comments on the TODOs. I also marked this as draft for now until it is ready for final review. It seems like the biggest blocker at the moment is the licensing of the `linux-audit-parser-rs` crate.

jorgehermo9:
Hi @jszwedko, I'm planning to address the review's comments now that the licensing issue was discussed. 

How can we address the feature-flag thing to comply with the license as you depicted in https://github.com/hillu/linux-audit-parser-rs/issues/1#issuecomment-2386577606? I've added it as an optional dependency in this PR and added to the stdlib feature flag https://github.com/vectordotdev/vrl/blob/95d695110d6643e9bde1d52779174f311a8ffe54/Cargo.toml#L87 but I don't think it is enough for what you said in that comment. What solution do you propose?

src/stdlib/parse_auditd.rs
jorgehermo9: https://github.com/hillu/linux-audit-parser-rs/pull/2

src/stdlib/parse_auditd.rs
jorgehermo9: https://github.com/hillu/linux-audit-parser-rs/pull/3

src/stdlib/parse_auditd.rs
```diff
@@ -0,0 +1,435 @@
...
+                    "pid" => 1_240_242,
+                    "uid" => 1_000,
+                    "auid" => 1_000,
+                    "ses" => 2,
+                    "msg"=> btreemap! {
+                        "op" => "PAM:accounting",
+                        "grantors" => "pam_unix,pam_permit,pam_time",
+                        "acct" => "vrl",
+                        "exe" => "/usr/bin/sudo",
+                        "hostname" => Value::Null,
```
jorgehermo9, line 421: Fields that are missing in the audit log (for example, `hostname="?"`) are parsed into `Null` values.

They are shown like this in vector:
```json
{
    "body": {
        "auid": 4294967295,
        "msg": {
            "addr": null,
            "comm": "systemd",
            "exe": "/usr/lib/systemd/systemd",
            "hostname": null,
            "res": "success",
            "terminal": null,
            "unit": "wpa_supplicant"
        },
        "pid": 1,
        "ses": 4294967295,
        "uid": 0
    },
    "enrichment": {
        "AUID": "unset",
        "UID": "root"
    },
    "sequence": 165,
    "timestamp": "2024-08-30T18:14:22.240Z",
    "type": "SERVICE_STOP"
}
```

Should we filter out that fields, or leave the entry with a null value, so people know that it is present but with no value?

src/stdlib/parse_auditd.rs
```diff
@@ -0,0 +1,435 @@
...
+        .collect::<Result<Vec<_>, _>>()?
+        .into_iter()
+        // Keys with only uppercase characters are considered enrichments of its lowercase counterpart
+        // https://github.com/linux-audit/audit-documentation/wiki/SPEC-Audit-Event-Enrichment
+        .partition(|(key, _)| key.chars().all(char::is_uppercase));
+
+    log.insert("body".into(), Value::from(body));
+
+    // TODO: should we downcase the keys of the enrichment so that they match its body counterparts?
+    if !enrichment.is_empty() {
```
jorgehermo9, line 68: In auditd, the association between fields and its enriched value is done with capital letters ([see this documentation](https://github.com/linux-audit/audit-documentation/wiki/SPEC-Audit-Event-Enrichment)).

But that limitation is because the log format is plain and it seems like a workaround.

As we can have a more structured representation and the enrichment fields are separated, maybe we can downcase the keys and have something like

(deleted some object keys to simplify the example)
```json
{
    "body": {
        "auid": 4294967295,
        "uid": 0
    },
    "enrichment": {
        "auid": "unset",
        "uid": "root"
    }
}
```

so people can link the body fields and the enrichment fields easily, as this way, they enriched counterparts will exactly match the raw keys.

What do you think? I don't really know how people would really consume this so I would appreciate some feedback from people who would actually use this (https://github.com/vectordotdev/vrl/issues/66 requesters)

src/stdlib/parse_auditd.rs
```diff
@@ -0,0 +1,435 @@
...
+            AuditValue::Owned(string) => Value::from(Bytes::from(string)),
+            AuditValue::Map(entries) => Value::from(
+                entries
+                    .into_iter()
+                    .map(|(key, value)| {
+                        Value::try_from(value).map(|value| (key.to_string().into(), value))
+                    })
+                    .collect::<Result<ObjectMap, _>>()?,
+            ),
+            // There are a few values that `linux-audit-parser` does not return in its parsing
```
jorgehermo9, line 97: depends on https://github.com/hillu/linux-audit-parser-rs/pull/4

src/stdlib/parse_auditd.rs
jorgehermo9: I don't really know...In the raw log lines this is shown as `a0=55b26d44a6a0` so maybe it makes sense to display it as `"a0": "0x55b26d44a6a0"` in my opinion. But I would love to know what requesters think about this...

jszwedko:
> Hi @jszwedko, I'm planning to address the review's comments now that the licensing issue was discussed.
> 
> How can we address the feature-flag thing to comply with the license as you depicted in [hillu/linux-audit-parser-rs#1 (comment)](https://github.com/hillu/linux-audit-parser-rs/issues/1#issuecomment-2386577606)? I've added it as an optional dependency in this PR and added to the stdlib feature flag
> 
> https://github.com/vectordotdev/vrl/blob/95d695110d6643e9bde1d52779174f311a8ffe54/Cargo.toml#L87
> 
> but I don't think it is enough for what you said in that comment. What solution do you propose?

Hey! Thanks for raising this again. After thinking about it some more, I'd prefer to avoid including LGPL dependencies in Vector, even as optional, since they run the risk of someone accidentally including them without realizing it (or us expanding usage without realizing it). I'd (unfortunately 🙁) prefer to see us find another dependency to parse auditd logs. We could even just leverage the existing support for grok parsing by using specific patterns to parse auditd logs. Example: https://gist.github.com/artbikes/2313040

jszwedko:
Apologies, I know this has been a bit of a long road to get to this point 😓 I'm curious what you think.

jorgehermo9:
Hi @jszwedko !

> I'd prefer to avoid including LGPL dependencies in Vector, even as optional, since they run the risk of someone accidentally including them without realizing it

I totally agree with that. It also was one of my main concerns about this...

>  prefer to see us find another dependency to parse auditd logs

I don't think there is something in rust yet. If I have enough time, I may end up writing my own parser with a MPL-compatible license. Maybe I can inspire from https://github.com/elastic/go-libaudit/blob/main/auparse/auparse.go as It is licensed as Apache2.0 and I think rewriting it (or inspiring from it) in rust could be feasible.

However, this may delay things a little bit, but I'll give it a try when I find enough time.

> Apologies, I know this has been a bit of a long road to get to this point 😓 I'm curious what you think.
Don't worry! Almost all VRL-related work is already done and it won't change a lot just by changing the underlying parser library. I think the same as you and I totally understand it.

Thanks! I may give a bump about this in a few weeks/month if I manage to implement the parser, but pausing it the moment.

jszwedko:
@jorgehermo9 thanks for understanding. I do think we might be able to get by with grok parsing instead of a native Rust parser. It'll be less efficient, but might be easier to get going. Worth considering anyway :)

jorgehermo9:
Hi @jszwedko I've working in a `Apache2.0 OR MIT` crate on my own, and will update this PR once I have my crate more polished (parsing auditd is **much harder**  than it may seem at a first glance).

Just to be sure, would this be OK from your side?

jszwedko:
> Hi @jszwedko I've working in a `Apache2.0 OR MIT` crate on my own, and will update this PR once I have my crate more polished (parsing auditd is **much harder** than it may seem at a first glance).
> 
> Just to be sure, would this be OK from your side?

Nice! That would be welcome.

edomaur:
I'm wondering, what's the status on this ?

jorgehermo9:
I'm still developing https://github.com/jorgehermo9/auditd-parser. I had to stop contributing to open source for a few months due to personal life... But I'm planning to start working in my parser library soon, and then including it here! parsing auditd is harder that it might seem, and I had to deep dive in some linux kernel internals, that's why it was taking me that much time...

So, I think I'll ship a basic version of the parser first. Because having the parser 100% valid is _very hard_, and only target regular linux kernel, as for example, some flavours such as rhel have different auditd log syntax in some cases... So -> my main focus is to land an initial 0.1 implementation of the parsing, for basic/most common logs, and then request users to fill issues with logs that the parser fails to parse, and work iteratively to gradually improve log coverage. Expect more news in the next ~2 months

PR #1011, opened 2024-09-01
fix(stdlib): make nullable timestamp and tags fields in `parse_influxdb` function

jorgehermo9:
Noticed that in `parse_influxdb` typedef it is not reflected that the `timestamp` and `tags` fields are nullable and may be missing when parsing.

Didn't add a changelog for this fix since I don't know how the typedef really affects the end user (and because that function is not really released yet)

### DIFF
--- a/src/stdlib/parse_influxdb.rs
+++ b/src/stdlib/parse_influxdb.rs
@@ -240,7 +240,7 @@ impl FunctionExpression for ParseInfluxDBFn {
 }
 
 fn tags_kind() -> Kind {
-    Kind::object(Collection::from_unknown(Kind::bytes()))
+    Kind::object(Collection::from_unknown(Kind::bytes())) | Kind::null()
 }
 
 fn gauge_kind() -> Kind {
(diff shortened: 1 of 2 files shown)


src/stdlib/parse_influxdb.rs
```diff
@@ -276,7 +276,7 @@ mod test {
         parse_influxdb => ParseInfluxDB;
 
         influxdb_valid {
-            args: func_args![ value: format!("cpu,host=A,region=us-west usage_system=64i,usage_user=10u,temperature=50.5,on=true,sleep=false 1590488773254420000") ],
+            args: func_args![ value: "cpu,host=A,region=us-west usage_system=64i,usage_user=10u,temperature=50.5,on=true,sleep=false 1590488773254420000" ],
```
jorgehermo9, line 279: Noticed this leftover `format!`

pront (approved):
Looks reasonable

jorgehermo9:
CI failed due to missing changelog @pront. Should I submit a changelog or is it not necessary for this fix?

jszwedko:
I think this should have a changelog entry. I added one since I'm prepping the next VRL release so this fix can be included.

jorgehermo9:
Thanks @jszwedko and sorry for the trouble!

jszwedko:
No trouble at all! Thanks for the fix!

PR #1015, opened 2024-09-04
enhancement(datadog-grok): extend `keyvalue` filter to match Datadog implementation

vladimir-dd:
Extends `keyvalue` filter to match Datadog [implementation](https://docs.datadoghq.com/service_management/events/pipelines_and_processors/grok_parser/?tab=filters#key-value-or-logfmt).

### DIFF
--- a/src/datadog/grok/filters/keyvalue.rs
+++ b/src/datadog/grok/filters/keyvalue.rs
@@ -117,10 +117,8 @@ fn regex_from_config(
     quotes: Vec<(char, char)>,
     _field_delimiters: (String, String),
 ) -> Result<Regex, GrokStaticError> {
-    let mut quoting = String::new();
-
     // start group
-    quoting.push('(');
+    let mut quoting = String::from("(");
     // add quotes with OR
     for (left, right) in quotes {
         quoting.push_str(&regex::escape(&left.to_string()));
--- a/src/datadog/grok/filters/keyvalue.rs
+++ b/src/datadog/grok/filters/keyvalue.rs
@@ -258,14 +258,16 @@ fn quoted(quotes: &[(char, char)]) -> impl Fn(&str) -> SResult<&str> + '_ {
 #[inline]
 fn parse_value<'a>(input: &'a str, quotes: &'a[(char, char)]) -> SResult<'a, Value> {
     alt((
-        map(quoted(quotes), |value| {
-            Value::Bytes(Bytes::copy_from_slice(value.as_bytes()))
-        }),
         parse_null,
         parse_boolean,
         parse_number,
-        rest.map(Value::from),
+        quoted(quotes).and_then(alt((parse_number, parse_string))),
+        parse_string,
     ))(input)
+}
+
+fn parse_string(input: &str) -> SResult<Value> {
+    map(rest, |s: &str| Value::Bytes(Bytes::copy_from_slice(s.trim().as_bytes())))(input)
 }
 
 fn parse_number(input: &str) -> SResult<Value> {
@@ -329,10 +331,17 @@ mod tests {
             Ok(("", Value::from("value"))),
             parse_value("value", DEFAULT_QUOTES)
         );
+        // trim whitespaces
+        assert_eq!(
+            Ok(("", Value::from("value"))),
+            parse_value(" value ", DEFAULT_QUOTES)
+        );
+        // remove quotes
         assert_eq!(
             Ok(("", Value::from("value"))),
             parse_value(r#""value""#, DEFAULT_QUOTES)
         );
+        // remove non-default quotes
         assert_eq!(
             Ok(("", Value::from("value"))),
             parse_value(r#"#value#"#, &[('#', '#')])
--- a/src/datadog/grok/filters/keyvalue.rs
+++ b/src/datadog/grok/filters/keyvalue.rs
@@ -82,12 +82,10 @@ pub fn filter_from_function(f: &Function) -> Result<GrokFilter, GrokStaticError>
             match f.args.as_ref().unwrap()[3] {
                 FunctionArgument::Arg(Value::Bytes(ref bytes)) => {
                     let delimiter_str = String::from_utf8_lossy(bytes);
-                    let mut delimiter_chars = delimiter_str.chars();
-                    match (delimiter_chars.next(), delimiter_chars.as_str()) {
-                        (Some(left), right) if !right.is_empty() => {
-                            (left.to_string(), right.to_string())
-                        }
-                        (Some(single), "") => (single.to_string(), single.to_string()),
+                    let mut chars = delimiter_str.chars();
+                    match (chars.next(), chars.next(), chars.as_str()) {
+                        (Some(single), None, _) => (single.to_string(), single.to_string()),
+                        (Some(left), Some(right), "") => (left.to_string(), right.to_string()),
                         _ => return Err(GrokStaticError::InvalidFunctionArguments(f.name.clone())),
                     }
                 }
--- a/src/datadog/grok/filters/keyvalue.rs
+++ b/src/datadog/grok/filters/keyvalue.rs
@@ -191,7 +191,19 @@ pub fn apply_filter(value: &Value, filter: &KeyValueFilter) -> Result<Value, Gro
                     || matches!(&value, Value::Bytes(b) if b.is_empty())
                     || key.is_empty())
                 {
-                    result.insert(crate::path!(key), value);
+                    let path = crate::path!(key);
+                    match result.get(path).cloned() {
+                        Some(Value::Array(mut values)) => {
+                            values.push(value);
+                            result.insert(path, values);
+                        }
+                        Some(prev) => {
+                            result.insert(path, Value::Array(vec![prev, value]));
+                        }
+                        None => {
+                            result.insert(path, value);
+                        }
+                    };
                 }
                 }
             });
--- a/src/datadog/grok/filters/keyvalue.rs
+++ b/src/datadog/grok/filters/keyvalue.rs
@@ -33,81 +33,81 @@ const DEFAULT_KEYVALUE_DELIMITER: &str = "=";
 const DEFAULT_QUOTES: &[(char, char)] = &[('"', '"'), ('\'', '\''), ('<', '>')];
 
 pub fn filter_from_function(f: &Function) -> Result<GrokFilter, GrokStaticError> {
-    {
-        let args_len = f.args.as_ref().map_or(0, |args| args.len());
+    let filter = key_value_filter_from_args(f.args.as_deref().unwrap_or_default().iter())
+        .ok_or_else(|| GrokStaticError::InvalidFunctionArguments(f.name.clone()))?;
+    Ok(GrokFilter::KeyValue(filter))
+}
 
-        let key_value_delimiter = if args_len > 0 {
-            match f.args.as_ref().unwrap()[0] {
-                FunctionArgument::Arg(Value::Bytes(ref bytes)) => &String::from_utf8_lossy(bytes),
-                _ => return Err(GrokStaticError::InvalidFunctionArguments(f.name.clone())),
-            }
-        } else {
-            DEFAULT_KEYVALUE_DELIMITER
-        };
-        let value_re = if args_len > 1 {
-            match f.args.as_ref().unwrap()[1] {
-                FunctionArgument::Arg(Value::Bytes(ref bytes)) => {
-                    [DEFAULT_VALUE_RE, &String::from_utf8_lossy(bytes)].concat()
+fn key_value_filter_from_args<'a>(
+    mut args: impl Iterator<Item = &'a FunctionArgument>,
+) -> Option<KeyValueFilter> {
+    let key_value_delimiter = match args.next() {
+        Some(FunctionArgument::Arg(Value::Bytes(ref bytes))) => &String::from_utf8_lossy(bytes),
+        Some(_) => return None,
+        None => DEFAULT_KEYVALUE_DELIMITER,
+    };
+
+    let value_re = match args.next() {
+        Some(FunctionArgument::Arg(Value::Bytes(ref bytes))) => {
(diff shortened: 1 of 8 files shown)


src/datadog/grok/filters/keyvalue.rs
```diff
@@ -125,27 +179,40 @@ pub fn apply_filter(value: &Value, filter: &KeyValueFilter) -> Result<Value, Gro
...
-            });
+            let value = String::from_utf8_lossy(bytes);
+            filter
+                .re_pattern
+                .captures_iter(value.as_ref())
+                .for_each(|c| {
+                    let key = parse_key(extract_capture(&c, 1), filter.quotes.as_slice());
+                    if key.contains(' ') {
+                        return;
+                    }
```
bruceg, line 190: Personally, I'd do this in the inverse just to avoid the early return, but TIOLI:
```suggestion
                    if !key.contains(' ') {
```

src/datadog/grok/filters/keyvalue.rs
```diff
@@ -125,27 +179,40 @@ pub fn apply_filter(value: &Value, filter: &KeyValueFilter) -> Result<Value, Gro
...
+                                }
+                                Some(prev) => {
+                                    result.insert(path, Value::Array(vec![prev, value]));
+                                }
+                                None => {
+                                    result.insert(path, value);
+                                }
+                            };
+                        }
+                    }
```
bruceg, line 214: I think this inner block is complex enough to warrant its own function. Is that reasonable? i.e. something like `.for_each(|c| filter.try_parse_value(&mut result, c))` or whatever other parameters it needs.

src/datadog/grok/filters/keyvalue.rs
```diff
@@ -302,84 +259,50 @@ fn quoted<'a>(
...
+    });
+    match res {
+        // check if it is a valid octal number(start with 0) - keep parsed as a decimal though
+        Ok((_, Value::Integer(_)))
+            if input.starts_with('0') && input.contains(|c| c == '8' || c == '9') =>
+        {
+            Err(nom::Err::Error((input, nom::error::ErrorKind::OctDigit)))
+        }
+        res => res,
     }
```
bruceg, line 306: Would this make sense to put above in the single branch that creates the `Value::Integer` variant? Also, does this secondary parsing now make this PR a breaking change, as the string `015` would parse as 15 before and 13 after the change?

src/datadog/grok/parse_grok.rs
```diff
@@ -255,29 +255,30 @@ mod tests {
     fn test_grok_pattern(tests: Vec<(&str, &str, Result<Value, Error>)>) {
         for (filter, k, v) in tests {
             let rules = parse_grok_rules(&[filter.to_string()], BTreeMap::new())
-                .expect("couldn't parse rules");
+                .unwrap_or_else(|_| panic!("failed to parse {k} with filter {filter}"));
```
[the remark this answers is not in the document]
vladimir-dd: Sure, might be useful.

src/datadog/grok/filters/keyvalue.rs
vladimir-dd: I prefer an early exit to be honest than having an extra nested level, however given that we can extract the code below into a function(as you suggest) this make sense.

src/datadog/grok/filters/keyvalue.rs
vladimir-dd: Good idea, updated https://github.com/vectordotdev/vrl/pull/1015/commits/401c44eca171b03f1c5ef47379cb622b46f0129e

src/datadog/grok/filters/keyvalue.rs
```diff
@@ -25,98 +25,152 @@ use super::super::{
...
             match f.args.as_ref().unwrap()[1] {
                 FunctionArgument::Arg(Value::Bytes(ref bytes)) => {
-                    let mut re_str = String::new();
-                    re_str.push_str(r"^[\w.\-_@");
+                    let mut re_str = String::from(DEFAULT_VALUE_RE);
                     re_str.push_str(&String::from_utf8_lossy(bytes));
-                    re_str.push_str("]+");
-                    Regex::new(re_str.as_str())
-                        .map_err(|_e| GrokStaticError::InvalidFunctionArguments(f.name.clone()))?
+                    re_str
```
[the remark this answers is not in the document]
vladimir-dd: Good idea - https://github.com/vectordotdev/vrl/pull/1015/commits/41ea1c9b6834d8faeb5481e37332948d149fbd33

src/datadog/grok/filters/keyvalue.rs
```diff
@@ -25,98 +25,152 @@ use super::super::{
...
-        };
+            Ok(Vec::from(DEFAULT_QUOTES))
+        }?;
 
         let field_delimiters = if args_len > 3 {
             match f.args.as_ref().unwrap()[3] {
                 FunctionArgument::Arg(Value::Bytes(ref bytes)) => {
-                    vec![String::from_utf8_lossy(bytes).to_string()]
+                    let delimiter = String::from_utf8_lossy(bytes).to_string();
+                    match (&delimiter[..1], &delimiter[1..]) {
```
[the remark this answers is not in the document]
vladimir-dd: >Also, [..1] and [1..] will only work if the first character is ASCII
Good catch, yeah, I remember running into this issue once.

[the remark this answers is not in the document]
vladimir-dd: Updated https://github.com/vectordotdev/vrl/pull/1015/commits/44d0801f7c39d5f4934771ea883728d4eca75041

src/datadog/grok/filters/keyvalue.rs
```diff
@@ -25,98 +25,152 @@ use super::super::{
...
+    // start group
+    quoting.push('(');
+    // add quotes with OR
+    for (left, right) in quotes {
+        quoting.push_str(&regex::escape(&left.to_string()));
+        quoting.push_str("[^");
+        quoting.push_str(&regex::escape(&left.to_string()));
+        quoting.push_str("]+");
+        quoting.push_str(&regex::escape(&right.to_string()));
+        quoting.push('|');
```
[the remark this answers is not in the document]
vladimir-dd: Good idea - https://github.com/vectordotdev/vrl/pull/1015/commits/48370561d9d70e9449ca286780a64f9fc01e95c6

src/datadog/grok/filters/keyvalue.rs
```diff
@@ -25,98 +25,152 @@ use super::super::{
...
+    keyvalue.push_str(key_value_delimiter);
+
+    // value
+    keyvalue.push_str(quoting.as_str());
+    keyvalue.push_str(&value_re);
+    keyvalue.push_str("]+)");
+
+    keyvalue.push_str("(?:[");
+    keyvalue.push_str(&_field_delimiters.1);
+    keyvalue.push_str("]|$)");
```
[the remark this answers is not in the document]
vladimir-dd: Good idea - https://github.com/vectordotdev/vrl/pull/1015/commits/d6edc7c6aa882b4b96f968249fb8fae0d7701b19

src/datadog/grok/filters/keyvalue.rs
vladimir-dd: Agree - extracted a function https://github.com/vectordotdev/vrl/pull/1015/commits/b4d67ad379fbb035d8c39c137bc8c894ab7fb1c8.
As to parsing octal numbers - we keep parsing them as decimal(no octal to decimal conversion), which corresponds to Datadog grok implementation.

src/datadog/grok/filters/keyvalue.rs
```diff
@@ -25,98 +26,152 @@ use super::super::{
...
                 }
                 _ => return Err(GrokStaticError::InvalidFunctionArguments(f.name.clone())),
             }
         } else {
-            // default field delimiters
-            vec![" ".into(), ",".into(), ";".into()]
+            (
+                DEFAULT_DELIMITERS.0.to_string(),
+                DEFAULT_DELIMITERS.1.to_string(),
+            )
```
bruceg, line 100: Final wondering, and this is in bike shed territory for sure: Maybe first assign the individual characters, and lift the repeated `.to_string()` calls out of the condition, or even leave them as `char`s when they are passed on.

src/datadog/grok/filters/keyvalue.rs
```diff
@@ -25,98 +26,152 @@ use super::super::{
...
+            key_value_delimiter: key_value_delimiter.to_string(),
         }))
     }
 }
 
+fn regex_from_config(
+    key_value_delimiter: &str,
+    value_re: String,
+    quotes: Vec<(char, char)>,
+    _field_delimiters: (String, String),
```
bruceg, line 120: nit: I'm curious why the underscore here? It is used below…

src/datadog/grok/filters/keyvalue.rs
```diff
@@ -155,145 +196,70 @@ pub fn apply_filter(value: &Value, filter: &KeyValueFilter) -> Result<Value, Gro
...
-    for &field_delimiter in field_delimiters {
-        match separated_list1(
-            parse_field_delimiter(field_delimiter),
-            parse_key_value(key_value_delimiter, field_delimiter, quotes, value_re),
-        )(input)
-        {
-            Ok((rest, v)) if rest.trim().is_empty() => {
-                return Ok((rest, v));
+fn parse_key_value_capture(
+    filter: &KeyValueFilter,
```
bruceg, line 200: Could this be a method on `KeyValueFilter`?

src/datadog/grok/filters/keyvalue.rs
```diff
@@ -155,145 +196,70 @@ pub fn apply_filter(value: &Value, filter: &KeyValueFilter) -> Result<Value, Gro
...
+                        values.push(value);
+                        result.insert(path, values);
+                    }
+                    Some(prev) => {
+                        result.insert(path, Value::Array(vec![prev, value]));
+                    }
+                    None => {
+                        result.insert(path, value);
+                    }
+                };
```
bruceg, line 227: The cloning here turns into some nasty O(N^2) behavior for array values, not to mention reinserting the same value at an existing slot. Could you use the `entry` interface to avoid needing to clone and reinsert?

src/datadog/grok/filters/keyvalue.rs
```diff
@@ -302,84 +268,57 @@ fn quoted<'a>(
...
+    map(rest, |s: &str| {
+        Value::Bytes(Bytes::copy_from_slice(s.trim().as_bytes()))
+    })(input)
+}
+
+fn parse_number(input: &str) -> SResult<Value> {
+    let res = map(terminated(double, eof), |v| {
+        if ((v as i64) as f64 - v).abs() == 0.0 {
+            // can be safely converted to Integer without precision loss
+            Value::Integer(v as i64)
```
bruceg, line 297: I was thinking to move the `validate_octal_number` bits right here so you don't have to re-match on the result value being an `Integer` since you already know it is. I see that doesn't work with the `map` function, is there another `nom` function that would allow returning an error?

bruceg (changes requested):
One blocking concern around array insertion behavior, otherwise just "this could be cleaner" comments.

src/datadog/grok/filters/keyvalue.rs
```diff
@@ -25,98 +26,152 @@ use super::super::{
...
-                        vec![(
-                            chars.next().expect("open quote"),
-                            chars.next().expect("closing quote"),
-                        )]
-                    } else if quotes.is_empty() {
-                        // default
-                        vec![('"', '"'), ('\'', '\''), ('<', '>')]
-                    } else {
-                        return Err(GrokStaticError::InvalidFunctionArguments(f.name.clone()));
+                    let pair = String::from_utf8_lossy(bytes);
```
tessneau, line 62: Not necessarily a blocker for this PR but -- Would it make sense to move this and the other similar sections, like the one retrieving the value for `field_delimiters`, into their own functions since they're pretty complex?

tessneau (approved):
looks good to me 👍  I'll defer to @bruceg for final review here since I'm not too familiar with this repo

bruceg:
This is ready for re-review. I'm not sure what to do about the errors in the benchmark code. It almost needs to be parsed from a VRL expression, but that's a significant overhaul compared to what's there.

pront:
> This is ready for re-review. I'm not sure what to do about the errors in the benchmark code. It almost needs to be parsed from a VRL expression, but that's a significant overhaul compared to what's there.

If it simplifies things, we can remove the benchmark to unblock this PR and add a ticket to re-introduce it as a followup.

pront (approved):
Thanks, this was a tricky one!

src/datadog/grok/filters/keyvalue.rs
```diff
@@ -156,72 +173,50 @@ fn regex_from_config(
...
-    match value {
-        Value::Bytes(bytes) => {
-            let mut result = Value::Object(BTreeMap::default());
-            let value = String::from_utf8_lossy(bytes);
-            filter
-                .re_pattern
-                .captures_iter(value.as_ref())
-                .for_each(|c| {
-                    parse_key_value_capture(filter, &mut result, c);
+impl KeyValueFilter {
```
tessneau, line 168: is the motivation behind having two seperate implementations organizational? i.e. first one concerning itself with initialization and this one with processing

tessneau:
Nicely done ! Thanks for closing the loop on this one

tessneau (approved):

src/datadog/grok/filters/keyvalue.rs
bruceg: Pretty much just minimizing the code movement in the diff to make it (a little) easier to review. I can move it if you think it's worth it.

tessneau: Ah gotcha 👍 feels fine as is, I just hadn't seen that much yet so I was mainly curious

PR #1022, opened 2024-09-06
fix(datadog grok): enable the DOTALL mode by default

vladimir-dd:
Enables the DOTALL mode(dot includes \n) for the Datadog grok lib by default.

### DIFF
--- a/changelog.d/1022.breaking.md
+++ b/changelog.d/1022.breaking.md
@@ -1,0 +1,2 @@
+The `parse_groks` VRL function has the multiline mode enabled by default.
+Use the `(?-m)` modifier to disable this behaviour.
--- a/changelog.d/1022.breaking.md
+++ b/changelog.d/1022.breaking.md
@@ -1,2 +1,2 @@
-The `parse_groks` VRL function has the multiline mode enabled by default.
+The multi-line mode of the `parse_groks` VRL function is now enabled by default.
 Use the `(?-m)` modifier to disable this behaviour.
--- a/src/datadog/grok/parse_grok_rules.rs
+++ b/src/datadog/grok/parse_grok_rules.rs
@@ -180,9 +180,13 @@ fn parse_pattern(
         // as opposed to the (?s) modifier in other regex flavors.
         // \A, \z - parses from the beginning to the end of string, not line(until \n)
         r"(?m)\A", // (?m) enables the DOTALL mode by default
-        &context.regex.replace("(?s)", "(?m)").replace("(?-s)", "(?-m)"),
+        &context
+            .regex
+            .replace("(?s)", "(?m)")
+            .replace("(?-s)", "(?-m)"),
         r"\z",
-    ].concat();
+    ]
+    .concat();
 
     // compile pattern
     let pattern = grok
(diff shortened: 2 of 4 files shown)


vladimir-dd:
Closing in favour of https://github.com/vectordotdev/vrl/compare/vladimir-dd/opa-2298-2

changelog.d/1022.breaking.md
```diff
@@ -0,0 +1,2 @@
+The `parse_groks` VRL function has the multiline mode enabled by default.
```
vladimir-dd, line 1: I'm not quite sure whether it makes sense to introduce an optional parameter for users to control this or simply keep existing behaviour for this function? 🤔  Datadog's grok implementation enabled multiline mode by default, which seems like a reasonable behaviour. Also users they can always use `(?m)` and `(?-m)` modifiers to control this behaviour.
cc @bruceg @jszwedko

src/datadog/grok/parse_grok_rules.rs
```diff
@@ -176,12 +176,13 @@ fn parse_pattern(
...
+    // In Oniguruma the (?m) modifier is used to enable the DOTALL mode(dot includes newlines),
+    // as opposed to the (?s) modifier in other regex flavors.
     // \A, \z - parses from the beginning to the end of string, not line(until \n)
-    pattern.push_str(r"\A");
+    pattern.push_str(r"(?m)\A"); // (?m) enables the DOTALL mode by default
     pattern.push_str(&context.regex);
     pattern.push_str(r"\z");
 
-    // our regex engine(onig) uses (?m) mode modifier instead of (?s) to make the dot match all characters
     pattern = pattern.replace("(?s)", "(?m)").replace("(?-s)", "(?-m)");
```
bruceg, line 186: nit: This could be done all in one go:
```rust
    let pattern = [
        // In Oniguruma the (?m) modifier is used to enable the DOTALL mode(dot includes newlines),
        // as opposed to the (?s) modifier in other regex flavors.
        // \A, \z - parses from the beginning to the end of string, not line(until \n)
        r"(?m)\A", // (?m) enables the DOTALL mode by default
        &context.regex.replace("(?s)", "(?m)").replace("(?-s)", "(?-m)"),
        r"\z",
    ].concat();
```

changelog.d/1022.breaking.md
bruceg, line 1:
```suggestion
The multi-line mode of the `parse_groks` VRL function is now enabled by default.
```

src/datadog/grok/parse_grok_rules.rs
vladimir-dd: Good idea, updated.

bruceg (approved):

changelog.d/1022.breaking.md
tessneau: I don't have as much context on `vrl` but my 2cents - agreed that making this the default behavior feels reasonable, and I think it would also make sense to eventually update `parse_grok` to have parity. If I'm looking at the [vrl docs](https://vector.dev/docs/reference/vrl/functions/#parse_grok) I would assume they share this base behavior.

PR #1031, opened 2024-09-09
fix(match_datadog_query): do not panic when attempting to parse invalid paths

### LINKED ISSUE #998: Panic when match_datadog_query references key with hyphen
### A note for the community

### Problem

`match_datadog_query` cannot handle keys with hyphens in them. For example:
```
.custom."abc_def" = true
exists = match_datadog_query(., "_exists_:@abc_def") #works as expected, returns true

.custom."abc-def" = true
exists = match_datadog_query(., "_exists_:@abc-def") # panics, regardless of it key exists
```

### Configuration

```text
Reproducible from VRL REPL
```

### Version

vector 0.40.0 (x86_64-apple-darwin 1167aa9 2024-07-29 15:08:44.028365803)

### Debug Output

```text
thread 'main' panicked at /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/vrl-0.17.0/src/stdlib/match_datadog_query.rs:418:42:
should parse value path (lookup_field): InvalidPathSyntax { path: "custom.abc-def" }
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: vrl::stdlib::match_datadog_query::lookup_field
   4: <vrl::stdlib::match_datadog_query::VrlFilter as vrl::datadog::filter::filter::Filter<vrl::value::value::Value>>::exists
   5: vrl::datadog::filter::matcher::build_matcher
   6: <vrl::stdlib::match_datadog_query::MatchDatadogQuery as vrl::compiler::function::Function>::compile
   7: vrl::compiler::compiler::Compiler::compile_function_call
   8: vrl::compiler::compiler::Compiler::compile_expr
   9: vrl::compiler::compile_with_state
  10: vrl::cli::repl::resolve
  11: <vrl::cli::repl::Repl as rustyline::validate::Validator>::validate
  12: rustyline::Editor<H,I>::readline_edit
  13: vrl::cli::cmd::run
  14: vrl::cli::cmd::cmd
  15: vector::cli::SubCommand::execute::{{closure}}
  16: vector::app::Application::prepare_from_opts
  17: vector::app::Application::run
  18: vector::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

### Example Data

_No response_

### Additional Context

_No response_

### References

_No response_

pront:
closes: https://github.com/vectordotdev/vrl/issues/998

[playground examples](https://playground.vrl.dev/?state=eyJwcm9ncmFtIjoiLnJlc3VsdDEgPSBtYXRjaF9kYXRhZG9nX3F1ZXJ5KHtcImFcIjogMX0sIFwiQGE6MVwiKVxuLnJlc3VsdDIgPSBtYXRjaF9kYXRhZG9nX3F1ZXJ5KHtcImEtYlwiOiBcIjFcIn0sIFwiQFxcXFxcXFwiYS1iXFxcXFxcXCI6MVwiKVxuLnRzID0gbm93KClcblxuLlwiYWJjX2RlZlwiID0gdHJ1ZVxuLmV4aXN0czEgPSBtYXRjaF9kYXRhZG9nX3F1ZXJ5KC4sIFwiX2V4aXN0c186YWJjX2RlZlwiKVxuXG4uXCJhYmMtZGVmXCIgPSB0cnVlXG4uZXhpc3RzMiA9IG1hdGNoX2RhdGFkb2dfcXVlcnkoLiwgXCJfZXhpc3RzXzpcXFxcXFxcImFiYy1kZWZcXFxcXFxcIlwiKSIsImV2ZW50Ijp7fSwiaXNfanNvbmwiOmZhbHNlLCJlcnJvciI6bnVsbH0%3D)

### DIFF
--- a/src/stdlib/match_datadog_query.rs
+++ b/src/stdlib/match_datadog_query.rs
@@ -1,5 +1,5 @@
 use std::borrow::Cow;
-
+use tracing::error;
 use crate::compiler::prelude::*;
 use crate::datadog_filter::{
     build_matcher,
@@ -120,8 +120,12 @@ impl Resolver for VrlFilter {}
 /// Implements `Filter`, which provides methods for matching against (in this case) VRL values.
 impl Filter<Value> for VrlFilter {
     fn exists(&self, field: Field) -> Box<dyn Matcher<Value>> {
-        let Ok(buf) = lookup_field(&field) else {
-            return Run::boxed(|_| false);
+        let buf = match lookup_field(&field) {
+            Ok(path) => path,
+            Err(error) => {
+                error!(message = "Invalid query field", %error);
+                return Run::boxed(|_| false);
+            }
         };
 
         match field {
@@ -157,9 +161,14 @@ impl Filter<Value> for VrlFilter {
     }
 
     fn equals(&self, field: Field, to_match: &str) -> Box<dyn Matcher<Value>> {
-        let Ok(buf) = lookup_field(&field) else {
-            return Run::boxed(|_| false);
+        let buf = match lookup_field(&field) {
+            Ok(path) => path,
+            Err(error) => {
+                error!(message = "Invalid query field", %error);
+                return Run::boxed(|_| false);
+            }
         };
+
 
         match field {
             // Default fields are compared by word boundary.
@@ -213,9 +222,14 @@ impl Filter<Value> for VrlFilter {
     }
 
     fn prefix(&self, field: Field, prefix: &str) -> Box<dyn Matcher<Value>> {
-        let Ok(buf) = lookup_field(&field) else {
-            return Run::boxed(|_| false);
+        let buf = match lookup_field(&field) {
+            Ok(path) => path,
+            Err(error) => {
+                error!(message = "Invalid query field", %error);
+                return Run::boxed(|_| false);
+            }
         };
+
 
         match field {
             // Default fields are matched by word boundary.
@@ -254,9 +268,14 @@ impl Filter<Value> for VrlFilter {
     }
 
     fn wildcard(&self, field: Field, wildcard: &str) -> Box<dyn Matcher<Value>> {
-        let Ok(buf) = lookup_field(&field) else {
-            return Run::boxed(|_| false);
+        let buf = match lookup_field(&field) {
+            Ok(path) => path,
+            Err(error) => {
+                error!(message = "Invalid query field", %error);
+                return Run::boxed(|_| false);
+            }
         };
+
 
         match field {
             Field::Default(_) => {
@@ -295,9 +314,14 @@ impl Filter<Value> for VrlFilter {
         comparator: Comparison,
         comparison_value: ComparisonValue,
     ) -> Box<dyn Matcher<Value>> {
-        let Ok(buf) = lookup_field(&field) else {
-            return Run::boxed(|_| false);
+        let buf = match lookup_field(&field) {
+            Ok(path) => path,
+            Err(error) => {
+                error!(message = "Invalid query field", %error);
+                return Run::boxed(|_| false);
+            }
         };
+
         let rhs = Cow::from(comparison_value.to_string());
 
         match field {
--- a/src/stdlib/match_datadog_query.rs
+++ b/src/stdlib/match_datadog_query.rs
@@ -5,11 +5,12 @@ use crate::datadog_filter::{
     Filter, Matcher, Resolver, Run,
 };
 use crate::datadog_search_syntax::{Comparison, ComparisonValue, Field, ParseError, QueryNode};
-use crate::diagnostic::Label;
 use crate::owned_value_path;
 use crate::path::{parse_value_path, OwnedValuePath, PathParseError};
-use crate::stdlib::match_datadog_query::Error::InvalidFilter;
 use std::borrow::Cow;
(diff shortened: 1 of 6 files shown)


src/stdlib/match_datadog_query.rs
```diff
@@ -120,7 +120,9 @@ impl Resolver for VrlFilter {}
 /// Implements `Filter`, which provides methods for matching against (in this case) VRL values.
 impl Filter<Value> for VrlFilter {
     fn exists(&self, field: Field) -> Box<dyn Matcher<Value>> {
-        let buf = lookup_field(&field);
+        let Ok(buf) = lookup_field(&field) else {
+            return Run::boxed(|_| false);
+        };
```
jszwedko, line 125: TIL this syntax works for assignments outside of conditions.

jszwedko (approved):
This is an improvement so I'm ok with seeing it merged, but should it be bubbling up the errors instead of silently ignoring the invalid fields?

pront:
> This is an improvement so I'm ok with seeing it merged, but should it be bubbling up the errors instead of silently ignoring the invalid fields?

Oh yes, good point. I added an `error!` for now. Propagating the error further will need some refactoring, I might follow-up in another PR.

src/stdlib/match_datadog_query.rs
```diff
@@ -120,7 +120,13 @@ impl Resolver for VrlFilter {}
...
 impl Filter<Value> for VrlFilter {
     fn exists(&self, field: Field) -> Box<dyn Matcher<Value>> {
-        let buf = lookup_field(&field);
+        let buf = match lookup_field(&field) {
+            Ok(path) => path,
+            Err(error) => {
+                error!(message = "Invalid query field", %error);
+                return Run::boxed(|_| false);
+            }
+        };
```
bruceg, line 129: This exact pattern is repeated several times here. Is there any way of DRYing it up other than a new macro?

bruceg (approved):

src/stdlib/match_datadog_query.rs
```diff
@@ -120,7 +120,13 @@ impl Resolver for VrlFilter {}
 /// Implements `Filter`, which provides methods for matching against (in this case) VRL values.
 impl Filter<Value> for VrlFilter {
     fn exists(&self, field: Field) -> Box<dyn Matcher<Value>> {
-        let buf = lookup_field(&field);
+        let buf = match lookup_field(&field) {
+            Ok(path) => path,
+            Err(error) => {
+                error!(message = "Invalid query field", %error);
+                return Run::boxed(|_| false);
```
jszwedko, line 127: I was thinking we'd change this function to return an error instead of logging it directly. Also, will this log for every event flowing through the system? It could flood, if so, in which case we should add `internal_log_rate_limit = true`.

jszwedko: Ah, I see your note about further refactoring. I'd be ok with just adding `internal_log_rate_limit = true` now if this would flood otherwise and refactor the function signature in a follow-up.

pront: Gotcha, I realize that it doesn't make sense to follow up in a second PR. I will update this PR.

src/stdlib/match_datadog_query.rs
pront: No longer relevant after https://github.com/vectordotdev/vrl/pull/1031/commits/551b0d21bc9521df08ef546fd557d7c3dc526573

src/stdlib/match_datadog_query.rs
pront: This commit https://github.com/vectordotdev/vrl/pull/1031/commits/551b0d21bc9521df08ef546fd557d7c3dc526573 eliminated all `error!` calls.

src/stdlib/match_datadog_query.rs
```diff
@@ -86,6 +88,38 @@ impl Function for MatchDatadogQuery {
...
+            Error::InvalidFilter(err) => err.fmt(f),
+        }
+    }
+}
+
+impl std::error::Error for Error {}
+
+impl DiagnosticMessage for Error {
+    fn code(&self) -> usize {
+        112
```
jszwedko, line 108: We could consider reusing the 403 error code ("invalid argument") for this: https://github.com/vectordotdev/vector/blob/master/website/cue/reference/remap/errors/403_invalid_argument.cue

src/stdlib/match_datadog_query.rs
```diff
@@ -86,6 +88,38 @@ impl Function for MatchDatadogQuery {
...
+    }
+
+    fn labels(&self) -> Vec<Label> {
+        match self {
+            Error::InvalidFilter(err) => {
+                vec![Label::primary(
+                    format!("grok pattern error: {err}"),
```
jszwedko, line 115: `grok`?

jszwedko (commented):
Nice! This is looking even better to me now. Just left a couple of comments inline.

src/stdlib/match_datadog_query.rs
pront: Done in https://github.com/vectordotdev/vrl/pull/1031/commits/6d01712ed53438a08464fed783734b4e61746ed0.

PR #1025, opened 2024-09-09
fix(stdlib): support hyphens in JIT parser

Lou-Kamades:
closes: #998

PR #1058, opened 2024-09-30
chore(releasing): Prepare VRL v0.19.0

pront:


CHANGELOG.md
```diff
@@ -4,6 +4,26 @@
...
+  Use the `(?-m)` modifier to disable this behaviour. (https://github.com/vectordotdev/vrl/pull/1022)
+
+### Enhancements
+
+- The `keyvalue` grok filter is extended to match Datadog implementation. (https://github.com/vectordotdev/vrl/pull/1015)
+
+### Fixes
+
+- The `parse_xml` function now doesn't add an unnecessary `text` key when processing single nodes. (https://github.com/vectordotdev/vrl/pull/849)
+- Adds support for hyphens in the JIT path parser (https://github.com/vectordotdev/vrl/pull/1025)
```
jszwedko, line 22: I'm not sure users will understand this one. What is the user-facing change?

jszwedko (commented):
Just one comment, otherwise 🚀

CHANGELOG.md
pront: This essentially means that now paths with a `-` e.g. `a-b-c` can be dynamically parsed while previously that would result in an error (or it would need quotes).

https://github.com/vectordotdev/vrl/pull/1025/files#diff-7b54bf8eddc8ffc8e619d38dcc2cd01afc4266b5bc6e3549d1b8e7877a72504e

Any suggestions here based on the above?

jszwedko: Ah I see, this just affects the grok parser? If so, I would say something like:

```
`parse_grok` and `parse_groks` no longer require field names containing a hyphen (e.g. `@a-b`) to be quoted.
```

jszwedko (approved):

PR #1097, opened 2024-10-28
chore(ast): Allow all `parser::ast` nodes to be built publically

gagbo:
- **Allow ast::Array and ast::Object to be built publicly**
- **Allow ast::Not nodes to be built publicly**

## Summary

In order to make extra tooling around VRL (mostly to be able to add preprocessing tooling on top of existing language features), it would be nice to be able to build all AST node types outside the crate.

Most of the nodes can be built, except those three, so I assume it’s fine to add support there. I do not know why the inner elements were made private for these, so instead of making those public, I added the "obvious" general constructors I could think of for them.

Regards,
Gerry

### DIFF
--- a/src/parser/ast.rs
+++ b/src/parser/ast.rs
@@ -538,6 +538,15 @@ impl IntoIterator for Array {
     }
 }
 
+impl FromIterator<Node<Expr>> for Array {
+    fn from_iter<T>(iter: T) -> Self
+    where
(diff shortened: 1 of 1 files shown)


pront (approved):
Thank you @gagbo, this PR makes sense to me.

> Most of the nodes can be built, except those three, so I assume it’s fine to add support there. I do not know why the inner elements were made private for these

It does seem that we have some accidental inconsistency here.

PR #1104, opened 2024-10-30
enhancement(punycode): improve performance for already coded cases

esensar:
This adds additional benchmarks for `encode_punycode` and `decode_punycode`, which demonstrate lower performance of these functions when `validate` flag is set to false. That is because without validation, a quick manual encoding / decoding is done, but it does not check beforehand if it is required at all.

This adds these checks for both encoding and decoding (and does so in case of decoding even when `validate` flag is `true`), which shows a big speedup in the benchmark for `validate: false` and a minor speedup for decoding with `validate: true`.

---

### DIFF
--- a/src/stdlib/decode_punycode.rs
+++ b/src/stdlib/decode_punycode.rs
@@ -71,6 +71,10 @@ impl FunctionExpression for DecodePunycodeFn {
         let value = self.value.resolve(ctx)?;
         let string = value.try_bytes_utf8_lossy()?;
 
+        if !string.contains(PUNYCODE_PREFIX) {
+            return Ok(string.into());
+        }
+
         let validate = self.validate.resolve(ctx)?.try_boolean()?;
 
         if validate {
--- a/src/stdlib/encode_punycode.rs
+++ b/src/stdlib/encode_punycode.rs
@@ -83,6 +83,10 @@ impl FunctionExpression for EncodePunycodeFn {
                 .map_err(|errors| format!("unable to encode to punycode: {errors}"))?;
             Ok(encoded.into())
         } else {
+            if string.is_ascii() {
+                return Ok(string.into());
+            }
+
             let encoded = string
                 .split('.')
                 .map(|part| {
--- a/benches/stdlib.rs
+++ b/benches/stdlib.rs
@@ -321,8 +321,18 @@ bench_function! {
         want: Ok("www.café.com"),
     }
 
+    encoded_no_validation {
+        args: func_args![value: "www.xn--caf-dma.com", validate: false],
+        want: Ok("www.café.com"),
+    }
+
     non_encoded {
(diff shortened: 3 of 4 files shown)


src/stdlib/decode_punycode.rs
```diff
@@ -71,6 +71,10 @@ impl FunctionExpression for DecodePunycodeFn {
         let value = self.value.resolve(ctx)?;
         let string = value.try_bytes_utf8_lossy()?;
 
+        if !string.contains(PUNYCODE_PREFIX) {
```
pront, line 74: I wonder if we can generally assume this.

esensar: I had trouble finding this in the punycode spec and then I remembered it is not in punycode spec, but it is used in IDN only, not in punycode in general. I think this might be safe to use, since by default this already validates if the value is a valid domain name (`validate: false` can be passed to skip that).

I could update the docs if it is not clear in the docs that this was the intended behavior.

src/stdlib/encode_punycode.rs
```diff
@@ -83,6 +83,10 @@ impl FunctionExpression for EncodePunycodeFn {
                 .map_err(|errors| format!("unable to encode to punycode: {errors}"))?;
             Ok(encoded.into())
         } else {
+            if string.is_ascii() {
```
pront, line 86: Same question ([as this](https://github.com/vectordotdev/vrl/pull/1104/files#r1831688371)) here.

src/stdlib/decode_punycode.rs
pront: Let's mention it in the docs.

src/stdlib/encode_punycode.rs
esensar: I think we can safely do it, since it is already done in the other branch. `idna` crate does a similar check right at the start (it also looks for the punycode prefix `xn--`, but that is only needed for validation and it is disabled in this branch), but I have noticed a slight difference - the library also normalizes uppercase characters to lowercase, so it only checks for lowercase characters.

I think I also need to fix this behavior now in general - to also replace uppercase character to lowercase when `validate: false`, for consistency with `validate: true` branch.

Should that fix be done here, or a separate PR?

pront: > Should that fix be done here, or a separate PR?
I would prefer a separate PR but both are acceptable.

pront (approved):
LGTM. There's a TIOLI [comment](https://github.com/vectordotdev/vrl/pull/1104/files#r1832890453) here.

src/stdlib/encode_punycode.rs
esensar: I think separate PR makes more sense too, so I will do it like that.

PR #1108, opened 2024-11-01
fix(convert): return error when conversion exceeds f64 bounds

### LINKED ISSUE #1107: to_float and infinity result can crash vector
We ran into an issue where parsing a string with `to_float` exceeds the bounds of `f64::MAX`, which returns a non-error infinity value.
```
$ str_value = "61687855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562"
"61687855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562"

$ parsed_value, err = to_float(str_value)
inf

$ err
null
```
Since `err` was `null`, it's might seem like it's safe to perform operations on this value but subtraction causes vector to panic and crash. This is because `f64::INFINITY - f64::INFINITY == f64::NAN` in rust.
```
$ parsed_value + parsed_value
inf

$ parsed_value + 1
inf

$ parsed_value - 1
inf

$ 1 - parsed_value
-inf

$ parsed_value - parsed_value
thread 'main' panicked at /Users/dan.hable/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ordered-float-4.2.2/src/lib.rs:1331:37:
Subtraction resulted in NaN: FloatIsNan
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
I can't find anything in the VRL language, besides using `to_string` to compare with "inf", that would allow me to detect this situation and avoid operations on infinite values. IMHO, I would have expected an error result instead of infinity, similar to how `to_int` behaves.
```
$ parsed_value, err = to_int(str_value)
"function call error for \"to_int\" at (20:29): Invalid integer \"61687855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562855628556285562\": number too large to fit in target type"
```

dhable:
Prior to this commit, string values that would exceed the f64 bounds would return `f64::INFINITY`. This leaks a type into the VRL scripts that could cause a panic without support to check for the value.

Fixes #1107

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Performance

## Is this a breaking change?

- [x] Yes
- [ ] No

## Checklist

### DIFF
--- a/changelog.d/1107.breaking.md
+++ b/changelog.d/1107.breaking.md
@@ -1,0 +1,1 @@
+Fixes the `to_float` function to return an error instead of `f64::INFINITY` when parsing [non-normal](https://doc.rust-lang.org/std/primitive.f64.html#method.is_normal) numbers.
--- a/src/compiler/conversion/mod.rs
+++ b/src/compiler/conversion/mod.rs
@@ -159,6 +159,11 @@ impl Conversion {
                 let parsed = s
                     .parse::<f64>()
                     .with_context(|_| FloatParseSnafu { s: s.clone() })?;
+                if !parsed.is_normal() {
+                    return Err(Error::NanFloat {
+                        s: format!("Invalid float \"{s}\": not a normal f64 number"),
+                    });
+                }
                 let f = NotNan::new(parsed).map_err(|_| Error::NanFloat { s: s.to_string() })?;
                 f.into()
             }
--- a/src/compiler/conversion/tests/mod.rs
+++ b/src/compiler/conversion/tests/mod.rs
@@ -2,7 +2,8 @@ use bytes::Bytes;
 use chrono::{DateTime, Utc};
 use ordered_float::NotNan;
 
-use crate::compiler::conversion::parse_bool;
+use crate::compiler::conversion::{parse_bool, Conversion, Error};
+use crate::compiler::TimeZone;
 
 #[cfg(unix)] // see https://github.com/vectordotdev/vector/issues/1201
 mod unix;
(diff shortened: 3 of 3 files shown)


changelog.d/1107.fix.md
```diff
@@ -0,0 +1,2 @@
+Fixes the `to_float` function to return an error instead of "infinity" when parsing a string outside
```
pront, line 1: Replace `fix` in the file name with `breaking`.

src/compiler/conversion/mod.rs
```diff
@@ -159,6 +159,11 @@ impl Conversion {
                 let parsed = s
                     .parse::<f64>()
                     .with_context(|_| FloatParseSnafu { s: s.clone() })?;
+                if parsed.is_infinite() {
```
pront, line 162: This is quite nuanced, but I think the least surprising behavior is to use https://doc.rust-lang.org/std/primitive.f64.html#method.is_normal here.

pront (commented):
Thanks @dhable, left a couple of comments but this change makes sense to me. 

Outside of this PR, I will take a look into the `to_float` vs `parse_float` implementations for `string` arguments.

pront:
Hi @dhable, thank you for this contribution. Do you plan to drive this PR to completion? Let me know if I can help.

dhable:
@pront Yes, I just got back from PTO but I'll get the suggestions added.

changelog.d/1107.breaking.md
```diff
@@ -0,0 +1,2 @@
+Fixes the `to_float` function to return an error instead of "infinity" when parsing a string outside
+the f64::MAX / f64::MIN bounds.
```
pront, line 2:
```suggestion
Fixes the `to_float` function to return an error instead of `f64::INFINITY` when parsing [non-normal](https://doc.rust-lang.org/std/primitive.f64.html#method.is_normal) numbers.
```

src/compiler/conversion/mod.rs
```diff
@@ -159,6 +159,11 @@ impl Conversion {
                 let parsed = s
                     .parse::<f64>()
                     .with_context(|_| FloatParseSnafu { s: s.clone() })?;
+                if !parsed.is_normal() {
+                    return Err(Error::NanFloat {
+                        s: format!("Invalid float \"{s}\": number too large to fit in target type"),
```
pront, line 164:
```suggestion
                        s: format!("Invalid float \"{s}\": not a normal f64 number"),
```

pront (approved):
Thank you @dhable. Left a couple of suggestions.

PR #1113, opened 2024-11-07
chore(parse_groks): expose internal errors

neuronull:
## Summary

`parse_grok` follows the Datadog Log's Grok algorithm's behavior where errors in applying filters are non fatal.
This change surfaces those filter errors to the caller, and differentiates between fatal and non fatal, as outlined by the Datadog Log's Grok algorithm, allowing the caller to decide whether to align with that or differ.

The stdlib `parse_groks` was unmodified, as such this is not a change to user facing behavior. If we decide to change that, we can do it in a follow up change.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## Is this a breaking change?

- [ ] Yes
- [x] No

## How did you test this PR?

Added/corrected unit tests

## Checklist

### DIFF
--- a/changelog.d/1113.breaking.md
+++ b/changelog.d/1113.breaking.md
@@ -1,0 +1,5 @@
+Currently if a grok filter as part of a `parse_groks` operation fails to apply due to an error against the input, the error is not surfaced and the field the filter applies to is not parsed, leading to data loss.
+
+This fix corrects that to surface errors while applying the grok filters.
+
+Existing parse_grok rules may need to be adjusted in order to correct filter queries that are now correctly surfaced as errors.
--- a/changelog.d/1113.breaking.md
+++ b/changelog.d/1113.breaking.md
@@ -1,5 +1,0 @@
-Currently if a grok filter as part of a `parse_groks` operation fails to apply due to an error against the input, the error is not surfaced and the field the filter applies to is not parsed, leading to data loss.
-
-This fix corrects that to surface errors while applying the grok filters.
-
-Existing parse_grok rules may need to be adjusted in order to correct filter queries that are now correctly surfaced as errors.
--- a/src/datadog/grok/parse_grok.rs
+++ b/src/datadog/grok/parse_grok.rs
@@ -9,7 +9,7 @@ use tracing::error;
 
 /// Errors which cause the Datadog grok algorithm to stop processing and not return a parsed result.
 #[derive(thiserror::Error, Debug, PartialEq, Eq)]
-pub enum FatalGrokError {
+pub enum FatalError {
     #[error("value does not match any rule")]
     NoMatch,
     #[error("failure during regex engine runtime for match of the pattern against the value.")]
@@ -18,7 +18,7 @@ pub enum FatalGrokError {
 
 /// Errors that do not prohibit the Datadog grok algorithm from continuing processing.
 #[derive(thiserror::Error, Debug, PartialEq, Eq)]
-pub enum RecoverableGrokError {
+pub enum InternalError {
     /// When this error is encountered, the value associated with the filter is not parsed into the
     /// resulting object.
     #[error("failed to apply filter '{}' to '{}'", .0, .1)]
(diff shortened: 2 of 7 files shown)


changelog.d/1113.breaking.md
```diff
@@ -0,0 +1,5 @@
+Currently if a grok filter as part of a `parse_groks` operation fails to apply due to an error against the input, the error is not surfaced and the field the filter applies to is not parsed, leading to data loss.
+
+This fix corrects that to surface errors while applying the grok filters.
+
+Existing parse_grok rules may need to be adjusted in order to correct filter queries that are now correctly surfaced as errors.
```
bruceg, line 5: Should this be marked as breaking (`!`) in the summary line?

neuronull: 👍  yes thanks for flagging. updated.

pront (approved):
Thanks, I agree that we should introduce this breaking change to fix the data loss scenario.

src/datadog/grok/parse_grok.rs
```diff
@@ -51,19 +51,15 @@ fn apply_grok_rule(source: &str, grok_rule: &GrokRule) -> Result<Value, Error> {
...
                 }) = grok_rule.fields.get(name)
                 {
-                    filters.iter().for_each(|filter| {
-                    if let Some(ref mut v) = value {
-                        value = match apply_filter(v, filter) {
-                            Ok(Value::Null) => None,
-                            Ok(v) if v.is_object() => Some(parse_keys_as_path(v)),
-                            Ok(v) => Some(v),
-                            Err(error) => {
-                                warn!(message = "Error applying filter", field = %field, filter = %filter, %error);
```
bruceg, line 61: Did we lose this warning when moving to the use of `?`? I don't see it in changes to callers.

src/datadog/grok/parse_grok.rs
```diff
@@ -51,19 +51,15 @@ fn apply_grok_rule(source: &str, grok_rule: &GrokRule) -> Result<Value, Error> {
...
-                                None
-                            }
-                        };
+                    for filter in filters {
+                        if let Some(ref mut v) = value {
+                            value = match apply_filter(v, filter)? {
+                                Value::Null => None,
+                                v if v.is_object() => Some(parse_keys_as_path(v)),
```
bruceg, line 58: I know this is copied code, but I think this could be written as the following to match the above:
```suggestion
                                Value::Object(_) => Some(parse_keys_as_path(v)),
```

src/datadog/grok/parse_grok.rs
neuronull: Oh I could call that out. The removal of the warning log was intentional. `parse_groks` stdlib function catches errors and logs them so when we change the behavior here to surface the error, this warning log is redundant.

src/datadog/grok/parse_grok.rs
```diff
@@ -51,19 +51,15 @@ fn apply_grok_rule(source: &str, grok_rule: &GrokRule) -> Result<Value, Error> {
...
-                            Ok(v) if v.is_object() => Some(parse_keys_as_path(v)),
-                            Ok(v) => Some(v),
-                            Err(error) => {
-                                warn!(message = "Error applying filter", field = %field, filter = %filter, %error);
-                                None
-                            }
-                        };
+                    for filter in filters {
+                        if let Some(ref mut v) = value {
+                            value = match apply_filter(v, filter)? {
```
tessneau, line 56: In the DD Logs parsing, if a filter fails, the rest of the parsed values are returned e.g. 
- input: `word not_an_integer another_word`
- rule `%{word:field1} %{data:field2:integer} %{word:field3}`
- result `{"field1": "word", "field3": "another_word"}`

I'm thinking that might be why we were returning `None` in the error case. If we return early in err I think we lose that, based on this unit test I used
```
#[test]
    fn continue_if_filter_fails() {
        test_full_grok(vec![
            // not a number
            (
                "%{word:field1} %{data:field2:integer} %{word:field3}",
                "word not_an_integer another_word",
                Ok(Value::from(btreemap! {
                    "field1" => Value::Bytes("word".into()),
                    "field3" => Value::Bytes("another_word".into()),
                })),
            ),
        ]);
    }
```

tessneau: btw for the Logs behavior I checked using a Grok Parser [here](https://dd.datad0g.com/logs/pipelines?search=team%3Aobservability_pipelines) in staging. I submitted some actual logs too but the sample testing UI seems reliable enough

neuronull: 👋  thanks for calling this out. I've adjusted the change to differentiate between fatal and non fatal errors to let the caller decide.

src/datadog/grok/parse_grok.rs
neuronull: I looked into it but changing the declaration like that doesnt work because the parse_keys_as_path expects to take ownership of the enum Value, not a variant of it.

tessneau (approved):
makes sense to me, thanks for the changes 🙏  should we remove the bug label with the latest update?

tessneau:
just noticed the bug label, should we remove it ?

src/datadog/grok/parse_grok.rs
```diff
@@ -5,37 +5,61 @@ use super::{
...
     NoMatch,
-    #[error("failure occurred during match of the pattern against the value: '{}'", .0)]
-    FailedToMatch(String),
+    #[error("failure during regex engine runtime for match of the pattern against the value.")]
+    RegexEngineError,
+}
+
+/// Errors that do not prohibit the Datadog grok algorithm from continuing processing.
+#[derive(thiserror::Error, Debug, PartialEq, Eq)]
+pub enum RecoverableGrokError {
```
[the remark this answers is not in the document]
pront: > I would argue that `NoMatch` is not actually an error

I agree, but I don't want to derail this PR.

> In what way are the latter errors "recoverable"? 

True, most VRL functions are expected to be idempotent.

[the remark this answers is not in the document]
neuronull: I agree the naming is weird but I would argue that it is somewhat unavoidable without larger changes.

I attempted to explain the "recoverable" errors in the doc comment, but I can adjust that if needed out of this thread's discussion.

So, "recoverable" here means the grok parser itself, not for the caller. That is why the "Fatal" is the only one returned in the `Result` of `parse_grok`.

I'm open to alternative naming suggestions than recoverable. I considered using "warning" but that seemed confusing because this is an `Error` type.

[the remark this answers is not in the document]
neuronull: i.e , these are errors that `parse_grok` itself recovered from.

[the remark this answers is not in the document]
bruceg: Maybe `Error` vs `Failure` might make sense, I don't know.

bruceg (approved):
I'm still confused by the error type naming, but I will approve this to unblock further work.

src/datadog/grok/parse_grok.rs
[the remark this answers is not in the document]
pront: > So, "recoverable" here means the grok parser itself, not for the caller.

Sounds like `internal` is a good qualification here.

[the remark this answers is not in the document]
neuronull: I'm on board for renaming `RecoverableGrokError` to `InternalGrokError` 👍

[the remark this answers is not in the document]
neuronull: > That makes sense to me as well. Do we need to repeat Grok in these type names?

👍 

and nope that is redundant, I'll remove it.

Thanks for that suggestion @pront

PR #1117, opened 2024-11-10
enhancement: Implement PartialOrd for Value

tontinton:
## Summary

Add `impl PartialOrd for Value`.

## Change Type

New feature.

## Is this a breaking change?

No.

## How did you test this PR?

Add unit test that calls `partial_cmp` with different values.

### DIFF
--- a/src/value/value.rs
+++ b/src/value/value.rs
@@ -3,7 +3,7 @@
 use bytes::{Bytes, BytesMut};
 use chrono::{DateTime, SecondsFormat, Utc};
 use ordered_float::NotNan;
-use std::collections::BTreeMap;
+use std::{cmp::Ordering, collections::BTreeMap};
 
 pub use iter::{IterItem, ValueIter};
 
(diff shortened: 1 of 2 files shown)


src/value/value.rs
```diff
@@ -172,6 +172,26 @@ impl Value {
...
+        match (self, other) {
+            (Value::Bytes(a), Value::Bytes(b)) => a.partial_cmp(b),
+            (Value::Regex(a), Value::Regex(b)) => a.partial_cmp(b),
+            (Value::Integer(a), Value::Integer(b)) => a.partial_cmp(b),
+            (Value::Float(a), Value::Float(b)) => a.partial_cmp(b),
+            (Value::Boolean(a), Value::Boolean(b)) => a.partial_cmp(b),
+            (Value::Timestamp(a), Value::Timestamp(b)) => a.partial_cmp(b),
+            (Value::Object(a), Value::Object(b)) => a.partial_cmp(b),
+            (Value::Array(a), Value::Array(b)) => a.partial_cmp(b),
+            (Value::Null, Value::Null) => Some(Ordering::Equal),
```
pront, line 189: The semantics here are quite tricky, for example in the SQL space `null =/= null`. If we are compared `null` variable in Java those would be true. I am tempted to say we should just remove this case and let it go to `None`.

tontinton: Left it as `None`, if someone has a better case for `null == null` we can change it later.

pront (approved):
Thanks @tontinton!

src/value/value.rs
```diff
@@ -172,6 +172,25 @@ impl Value {
...
+            return None;
+        }
+        match (self, other) {
+            (Self::Bytes(a), Self::Bytes(b)) => a.partial_cmp(b),
+            (Self::Regex(a), Self::Regex(b)) => a.partial_cmp(b),
+            (Self::Integer(a), Self::Integer(b)) => a.partial_cmp(b),
+            (Self::Float(a), Self::Float(b)) => a.partial_cmp(b),
+            (Self::Boolean(a), Self::Boolean(b)) => a.partial_cmp(b),
+            (Self::Timestamp(a), Self::Timestamp(b)) => a.partial_cmp(b),
+            (Self::Object(a), Self::Object(b)) => a.partial_cmp(b),
```
pront, line 187: I wonder what's the behavior here.

Issue #997, opened 2024-08-20
support for some  hash function like crc32


geekfirstty:
I want to use crc32 hash function in remap, can you provide support for this.

jszwedko:
I think this would be a reasonable addition. If anyone wants to take a shot at this the implementation of the `md5` function should be a good reference: https://github.com/vectordotdev/vrl/blob/main/src/stdlib/md5.rs

(I moved this issue to the VRL repository since it is asking for a new function)

ivor11:
Hi @jszwedko ,
I've raised a PR for this issue
https://github.com/vectordotdev/vrl/pull/1136

jszwedko:
Agreed, closed by #1136

PR #1136, opened 2024-11-22
feat(crc): add crc hash fn

ivor11:
## Summary

Add crc hash fn with an algorithm parameter
default algorithm used - CRC_32_ISO_HDLC

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Performance

## Is this a breaking change?

- [ ] Yes
- [x] No

## Checklist

## References

https://github.com/vectordotdev/vrl/issues/997


src/stdlib/crc32.rs
```diff
@@ -0,0 +1,77 @@
...
+
+impl Function for Crc32 {
+    fn identifier(&self) -> &'static str {
+        "crc32"
+    }
+
+    fn parameters(&self) -> &'static [Parameter] {
+        &[Parameter {
+            keyword: "value",
+            kind: kind::ANY,
```
pront, line 22: Hm I noticed this in other hash function. I don't think `ANY` is ideal but we can leave it here for consistency.

pront (approved):
Thank you @ivor11. This looks good. We will need a PR on Vector to document this new function.

ivor11:
Awesome @pront. I have created a PR on vector to document this function
https://github.com/vectordotdev/vector/pull/21924

src/stdlib/crc32.rs
bruceg, line 1: There are many kinds of CRC functions, even many kinds of 32-bit CRC functions. Instead of adding a separate function for each, would it make sense to instead have an "algorithm" parameter like `decrypt`, `encrypt`, and `hmac` use?

pront: Good point. @ivor11 should be trivial to modify this PR to do the above. Let me know if this makes sense.

ivor11: hey @bruceg @pront , 
I've added the algorithm parameter - keeping CRC_32_ISO_HDLC as the default which I suppose is the most popular one

pront: Awesome, thanks. I will take a look at the updated code.

src/stdlib/crc.rs
```diff
@@ -0,0 +1,587 @@
+use crate::compiler::prelude::*;
+use crate::value;
+use crc::Crc as CrcInstance;
+
+fn crc(value: Value, algorithm: Value) -> Resolved {
```
pront, line 5: 💭 This is one of those (rare) cases where a `#[proc_macro]` can significantly reduce the boilerplate. We could iterate over `valid_algorithms` and keep pushing:
```
 $algo => CrcInstance::<u8>::new(&crc::$algo)
            .checksum(&value)
            .to_string(),
```

But this is just my observation. Feel free to ignore this comment.

src/stdlib/crc.rs
```diff
@@ -0,0 +1,587 @@
...
+            tdef: TypeDef::bytes().infallible(),
+        }
+
+        crc_crc64 {
+            args: func_args![value: "foo", algorithm: "CRC_64_REDIS"],
+            want: Ok(value!(b"12626267673720558670")),
+            tdef: TypeDef::bytes().infallible(),
+        }
+    ];
+}
```
pront, line 587: can you add a test case with an invalid algorithm?

src/stdlib/crc.rs
```diff
@@ -0,0 +1,587 @@
...
+        let algorithm = match &self.algorithm {
+            Some(algorithm) => algorithm.resolve(ctx)?,
+            None => value!("CRC_32_ISO_HDLC"),
+        };
+
+        crc(value, algorithm)
+    }
+
+    fn type_def(&self, state: &state::TypeState) -> TypeDef {
+        let valid_algorithms = [
```
pront, line 421: Let's move this at the top as a `const VALID_ALGORITHMS: &[&str] = &[`

src/stdlib/crc.rs
```diff
@@ -0,0 +1,587 @@
...
+            "CRC_64_ECMA_182",
+            "CRC_64_GO_ISO",
+            "CRC_64_MS",
+            "CRC_64_REDIS",
+            "CRC_64_WE",
+            "CRC_64_XZ",
+            "CRC_82_DARC",
+        ];
+
+        let mut valid_static_algo = false;
```
pront, line 536: I think we can simplify this a little bit:
```
   fn type_def(&self, state: &state::TypeState) -> TypeDef {
        if let Some(algorithm) = self.algorithm.as_ref() {
            if let Some(algorithm) = algorithm.resolve_constant(state) {
                if let Ok(algorithm) = algorithm.try_bytes_utf8_lossy() {
                    let algorithm = algorithm.to_uppercase();
                    if VALID_ALGORITHMS.contains(&algorithm.as_str()) {
                        return TypeDef::bytes().infallible();
                    }
                }
            }
        } else {
            return TypeDef::bytes().infallible();
        }

        TypeDef::bytes().fallible()
    }
```
(as long as the tests are passing)

pront (approved):

pront (approved):

pront:
This also need a changelog to communicate this new feature to the users.

ivor11:
@pront my bad, I've added the changelog file. Thanks for the thorough review

Issue #1157, opened 2024-12-02
Add functions to map an array into an object


bruceg:
When manipulating data that has been parsed from CSV or other columnar data sources, it is useful to be able to map those columns into an object with named fields. I propose adding two functions to accomplish this, mirroring the approach in Python: `object` and `zip`:

* `zip` takes a number of arrays as input and produces a single array as output containing a tuple of elements from each array. For example `zip([1,2,3],[4,5,6],[7,8,9])` returns `[[1,4,7],[2,5,8],[3,6,9]]`.
* `object` takes takes a zipped array as above with (key, value) pairs and produces an object as output. For example `object([["foo":1],["bar":2]])` returns `{"foo":1,"bar":2}`.

jszwedko:
I'd suggest a different name for the function that takes an array of "tuples" than `object`, which [already exists](https://vector.dev/docs/reference/vrl/functions/#object). Maybe `from_pairs`? That's what lodash calls it. Maybe there are other examples from other languages or libraries we could use for inspiration.

bruceg:
I missed that `object` was already a function, and I'm on board with something like `from_pairs`, but isn't that a little ambiguous as to what the _output_ of such a function would be?

jszwedko:
True, could call it `object_from_pairs` to be explicit. Granted we have other functions already where the return type isn't clear just from the name itself (e.g. [from_unix_timestamp](https://vector.dev/docs/reference/vrl/functions/#from_unix_timestamp)).

PR #1164, opened 2024-12-03
feat(stdlib): Add new `object_from_array` function

bruceg:
## Summary

Adds a new `object_from_array` function as discussed in #1157 

## Change Type

- [ ] Bug fix
- [X] New feature
- [ ] Performance

## Is this a breaking change?

- [ ] Yes
- [x] No

## How did you test this PR?

Unit tests are included

- [X] Yes. Please add a changelog fragment based on

## Checklist

### DIFF
--- a/src/stdlib/object_from_array.rs
+++ b/src/stdlib/object_from_array.rs
@@ -1,0 +1,130 @@
+use crate::compiler::prelude::*;
+
+fn make_object(values: Vec<Value>) -> Resolved {
+    values
+        .into_iter()
+        .map(make_key_value)
+        .collect::<Result<_, _>>()
+        .map(Value::Object)
+}
+
+fn make_key_value(value: Value) -> Result<(KeyString, Value), ExpressionError> {
+    value.try_array().map_err(Into::into).and_then(|array| {
+        let mut iter = array.into_iter();
+        let key: KeyString = match iter.next() {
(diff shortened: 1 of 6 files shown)


bruceg:
Note that I plan a follow-up PR to add support for an optional `keys` parameter to avoid the need for `zip`.

Also, I don't really like the name `object_from_array` and would love a better suggestion. If you think `from_array` actually makes more sense, I can do that too.

bruceg:
Also, this duplicates the `enum ConstOrExpr` type from `zip`. Where would that best live in common code?

src/stdlib/object_from_array.rs
```diff
@@ -0,0 +1,130 @@
...
+        make_object(self.values.resolve(ctx)?.try_array()?)
+    }
+
+    fn type_def(&self, _state: &TypeState) -> TypeDef {
+        TypeDef::object(Collection::any())
+    }
+}
+
+#[derive(Clone, Debug)]
+enum ConstOrExpr {
```
bruceg, line 76: This is a straight-up clone of the construct in `zip`. Where would it make sense to move it into common code?

pront: Yes, we can introduce a `src/stdlib/utils.rs` (for the lack of a better name)

pront (approved):
Thanks!

PR #1179, opened 2024-12-06
fix(convert): Revert PR #1108 "return error when conversion exceeds f64 bounds"

pront:
## Summary

Reverts vectordotdev/vrl#1108
Due to https://github.com/vectordotdev/vector/issues/21970. We don't have numbers to prove this, but parsing `"0"` might be a common operation. And we don't want `parse_float` and `to_float` to diverge.

To err on the side of caution let's revert this change. We will work to provide safe arithemtic operations in the future (inspired by Rust APIs).

### DIFF
--- a/changelog.d/1179.fix.md
+++ b/changelog.d/1179.fix.md
@@ -1,0 +1,2 @@
+Reverted `to_float` [change](https://github.com/vectordotdev/vrl/pull/1107) because the new logic is too restrictive
+e.g. attempting to convert "0" returns an error.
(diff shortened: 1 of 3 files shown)


pront:
Do we usually add a fix changelog in these cases? cc @jszwedko

jszwedko (approved):

jszwedko:
> Do we usually add a fix changelog in these cases? cc @jszwedko

Yeah, I think a `fix` changelog is right here.

PR #1197, opened 2024-12-27
feat(stdlib): allow parsing multiple durations at once in `parse_duration` function

titaneric:
## Summary

Hi, I hope that `parse_duration` could parse multiple duration at once, so I have opened #1198 as a referenced issue.

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Non-functional (chore, refactoring, docs)
- [ ] Performance

## Is this a breaking change?

- [ ] Yes
- [x] No

## How did you test this PR?

```
cargo test
./scripts/check.sh
```

## Does this PR include user facing changes?

- [ ] Yes. Please add a changelog fragment based on
  our [guidelines](https://github.com/vectordotdev/vector/blob/master/changelog.d/README.md).
- [x] No. A maintainer will apply the "no-changelog" label to this PR.

## Checklist

- [x] Our [CONTRIBUTING.md](https://github.com/vectordotdev/vrl/blob/main/CONTRIBUTING.md) is a good starting place.
- [x] If this PR introduces changes to [LICENSE-3rdparty.csv](https://github.com/vectordotdev/vrl/blob/main/LICENSE-3rdparty.csv), please
  run `dd-rust-license-tool write` and commit the changes. More details [here](https://crates.io/crates/dd-rust-license-tool).
- [x] For new VRL functions, please also create a sibling PR in Vector to document the new function.

## References

Closes #1196

PR #1223, opened 2025-01-21
fix(stdlib): human time parse duration error for decimals

### LINKED ISSUE #22255: Vector 0.44.0 no longer support decimal in parse_duration
### A note for the community

_No response_

### Problem

Before 0.44.0, we could use parse_duration to parse "12.3s" to 12.3.

After upgrading to 0.44.0, this resulted in an error `unable to parse duration: 'invalid character at 2'`

Likely due to the change in https://github.com/vectordotdev/vrl/pull/1197 which should have been listed in the upgrade guide.

### Configuration

```text

```

### Version

0.44.0

### Debug Output

```text

```

### Example Data

_No response_

### Additional Context

_No response_

### References

_No response_

sainad2222:
Credits to @titaneric for [original](https://github.com/vectordotdev/vrl/pull/1197) implementation

## Summary

New humantime parse duration function doesn't support decimals. So falling back to old method(regex) when ht_parse_duration fails

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Performance

## Is this a breaking change?

- [ ] Yes
- [x] No

## How did you test this PR?

```
cargo test parse_duration
./scripts/checks.sh
```

## Checklist

## References

- Closes: https://github.com/vectordotdev/vector/issues/22255


titaneric:
Thank you for trying for fixing this! Is it possible to support multi-units duration in the implementation as well?
In fact, I have the [previous implementation](https://github.com/titaneric/vrl/blob/1dbb35fe10d85bc06b6d8336e1abca8bf0f6f00f/src/stdlib/parse_duration.rs#L7) here.

pront:
> Thank you for trying for fixing this! Is it possible to support multi-units duration in the implementation as well? In fact, I have the [previous implementation](https://github.com/titaneric/vrl/blob/1dbb35fe10d85bc06b6d8336e1abca8bf0f6f00f/src/stdlib/parse_duration.rs#L7) here.

Good idea, it could have been in a separate function but I am fine to include in this one as long as we include some more tests 👍

titaneric:
Perhaps testcase for multi-unit and fractional time duration?

sainad2222:
Ahh got it

sainad2222:
Added. Thanks @titaneric for the logic

jszwedko:
@titaneric if we are adding support to the existing implementation to handle multiple units, is there any reason to keep the implementation using `humantime` around? Does it offer any other advantages that would motivate supporting both implementations?

titaneric:
I am not sure, to be honest.
I think if we need to handle fractional and multi-units duration parsing, we need to handle by ourselves.

I think we could remove the humantime crate this time.

src/stdlib/parse_duration.rs
```diff
@@ -206,31 +229,59 @@ mod tests {
...
+            want: Err("unable to parse duration: '1'"),
             tdef: TypeDef::float().fallible(),
         }
 
-        error_format {
+        s_w {
             args: func_args![value: "1s",
                              unit: "w"],
-            want: Err("unknown unit format: 'w'"),
+            want: Ok(0.000_001_653_439_153_439_153_5),
```
sainad2222, line 277: Not sure if we want to support weeks. Either we remove support for both w_ns and s_w or keep it as it is. Let me your thoughts

titaneric: I think week is a common unit, and I think we could support it

titaneric (commented):
I think it's ok, and I would leave this PR to maintainer to approve it.

src/stdlib/parse_duration.rs
pront: Thanks for supporting a week here. Maybe we can have another test as well e.g. "8d"?

sainad2222: Added

pront (approved):
Awesome, thank you @sainad2222 and @titaneric.

PR #1270, opened 2025-02-13
chore(dev): more clippy fixes

pront:
## Summary

Continuing the cleanup efforts.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## Is this a breaking change?

- [ ] Yes
- [x] No

## Checklist

## References

Previous PR: https://github.com/vectordotdev/vrl/pull/1259

### DIFF
--- a/src/datadog/filter/matcher.rs
+++ b/src/datadog/filter/matcher.rs
@@ -99,8 +99,7 @@ where
 /// # Errors
 ///
 /// Will return `Err` if the query contains an invalid path.
-#[allow(clippy::module_name_repetitions)]
-pub fn build_matcher<V, F>(
+pub fn build<V, F>(
     node: &QueryNode,
     filter: &F,
 ) -> Result<Box<dyn Matcher<V>>, PathParseError>
@@ -196,11 +195,11 @@ where
 
             Ok(any(matchers?))
         }
-        QueryNode::NegatedNode { node } => Ok(not(build_matcher(node, filter)?)),
+        QueryNode::NegatedNode { node } => Ok(not(build(node, filter)?)),
         QueryNode::Boolean { oper, nodes } => {
             let funcs: Result<Vec<_>, _> = nodes
                 .iter()
-                .map(|node| build_matcher(node, filter))
+                .map(|node| build(node, filter))
                 .collect();
 
             match oper {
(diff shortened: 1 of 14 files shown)


src/value/value/lua.rs
```diff
@@ -107,10 +107,19 @@ pub fn table_to_timestamp(t: LuaTable) -> LuaResult<DateTime<Utc>> {
...
-        .and_hms_nano_opt(hour, min, sec, nano)
-        .expect("invalid timestamp"))
+
+    let base_dt = Utc
+        .with_ymd_and_hms(year, month, day, hour, min, sec)
+        .single()
+        .ok_or_else(|| mlua::Error::external("invalid timestamp"))?;
+
+    if nano > 0 {
+        base_dt
```
pront, line 117: This was the only change that needed some thought.

src/datadog/filter/matcher.rs
```diff
@@ -99,6 +99,7 @@ where
 /// # Errors
 ///
 /// Will return `Err` if the query contains an invalid path.
+#[allow(clippy::module_name_repetitions)]
 pub fn build_matcher<V, F>(
```
jszwedko, line 103: I kinda agree the method name is redundant here. We could just call this `build`.

src/datadog/filter/regex.rs
```diff
@@ -4,6 +4,7 @@ use regex::Regex;
 ///
 /// # Panics
 /// Panics if an invalid wildcard regex is provided.
+#[allow(clippy::module_name_repetitions)]
 #[must_use]
 pub fn word_regex(to_match: &str) -> Regex {
```
jszwedko, line 9: Same here, could just call this `word` and the below `wildcard` given the module is called `regex`.

jszwedko (approved):
Left a few nits, otherwise 👍 Happy to see more clippy lints being taken care of.

src/datadog/filter/matcher.rs
jszwedko: Ah, good point. That might be worth leaving a comment to indicate why we are ignoring the clippy lint.

If we wanted to, we could go further to introduce the new methods with new names and deprecate these ones.

PR #1114, opened 2024-11-07
fix(encode_proto): When encoding strings into numeric proto fields, attempt to convert them automatically

flavioc:
## Summary
Before this change, attempting to convert a VRL value that contained bytes into a numeric proto field would fail since we expect the VRL value to be numeric too. This PR makes `encode_proto` automatically convert those strings into numeric values automatically. We still fail if the format is not valid.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Non-functional (chore, refactoring, docs)
- [ ] Performance

## Is this a breaking change?

- [ ] Yes
- [x] No

## Does this PR include user facing changes?

- [x] Yes. Please add a changelog fragment based on
  our [guidelines](https://github.com/vectordotdev/vector/blob/master/changelog.d/README.md).
- [ ] No. A maintainer will apply the "no-changelog" label to this PR.

## Checklist

- [x] Our [CONTRIBUTING.md](https://github.com/vectordotdev/vrl/blob/main/CONTRIBUTING.md) is a good starting place.
- [x] If this PR introduces changes to [LICENSE-3rdparty.csv](https://github.com/vectordotdev/vrl/blob/main/LICENSE-3rdparty.csv), please
  run `dd-rust-license-tool write` and commit the changes. More details [here](https://crates.io/crates/dd-rust-license-tool).
- [x] For new VRL functions, please also create a sibling PR in Vector to document the new function.

PR #1304, opened 2025-03-01
enhancement(encode_proto): When encoding values into string fields, attempt to convert them automatically

roykim98:
## Summary
Similar to https://github.com/vectordotdev/vrl/pull/1114, encode values to string fields by converting values of `bool`, `int`, and `float` under the hood. This is an operation that fails currently.

Also decodes `test.desc` such that a `test.proto` is included alongside the other existing `test*.proto` and `test*.desc` in case it needs to be updated in the future.

## Change Type

- [x] Bug fix
- [x] New feature
- [ ] Performance

## Is this a breaking change?

- [ ] Yes
- [x] No

## How did you test this PR?

Unit tests

## Checklist


changelog.d/1304.fix.md
```diff
@@ -0,0 +1 @@
+The `encode_proto` function was enhanced to automatically convert integer, float, and boolean values when passed to string proto fields.
```
pront, line 1: Filename: Enhancement is a bit more accurate here.

tests/data/protobuf/test.proto
```diff
@@ -0,0 +1,61 @@
+syntax = "proto3";
```
pront, line 1: Does this map to the existing `tests/data/protobuf/test.desc`? Optional, but it would be nice to introduce something like this https://github.com/vectordotdev/vector/pull/22536/files#diff-8a0bd3036d2fcba538b9767c3858583db8b5e239df796618fa79f6302801f62f.

pront (approved):
Thank you @roykim98. This makes sense 👍

tests/data/protobuf/test.proto
roykim98: @pront 
Yes, I reverse-engineered the original input, and then regenerated a new `test.desc` and ran the tests with it and found that they passed.

Can you clarify what you're asking me to do? I agree having a README and a makefile would be helpful, but it looks like that's already being covered in an open PR in the link you provided.

roykim98: Oh, I see. Are you asking that I create the `Makefile` for the existing proto files such that https://github.com/vectordotdev/vector/pull/22536/files#diff-4eaacf426d6dbe58dbeb003a6cf12b91aadc41930cb664135fee29a7db5d5b18 only needs to add an incremental line for the new proto file it introduces?

pront: > Yes, I reverse-engineered the original input
Awesome, I was wondering if you did that, thanks :)

> only needs to add an incremental line for the new proto file it introduces?

Correct. I think you can copy/paste that file and change the proto paths.

roykim98: Think I would need to add some kind of fetching logic to also include `google/protobuf/Timestamp.proto`, but yeah I can take a look.

roykim98: Added and regenerated `./scripts/check.sh` passes. Not sure why `test_protobuf.desc` and `test_protobuf3.desc` has a diff, maybe compiled with a different version. But the tests pass.

pront: Thank you @roykim98

Issue #1095, opened 2024-10-25
JIT compilation


tontinton:
Wondering if there are plans to compile into machine code.

pront:
Hi @tontinton,

We currently have a REPL implementation at https://github.com/vectordotdev/vrl/blob/main/src/cli/repl.rs.

Would you mind sharing what use cases you're thinking about? This would help us understand if you're envisioning something different.

For context, you might find these resources helpful:

- https://vector.dev/docs/reference/vrl/#goals
- https://vector.dev/docs/reference/vrl/#vector-rust-native

tontinton:
I want to run VRL on lots and lots of logs (Terabytes) for implementing a project similar to Trino but for semi-structured logs.

I am concerned about the performance of VRL as an interpreted language. LuaJIT and CRuby for example implement JIT compilation, and wondered whether this is something VRL should also have?

pront:
Gotcha, there is no plan at the moment to do this. We will update this ticket if this changes in the future.

Out of curiosity, did you run any benchmarks? You can write a very simple Rust crate which uses VRL as a library. You can see [here](https://github.com/vectordotdev/vector/blob/efc9db951f144d56f2f6b0182f5659661d79a082/src/transforms/remap.rs#L208-L210) how Vector's `remap` transform does it.

tontinton:
No benchmarks yet, but planning to.

PR #1305, opened 2025-03-01
[DRAFT] Performance - Remove unneeded value clones

tontinton:
I ran a pretty simple VRL that looks like:

```py
result = []
t = now()
for_each(array!(.SomeList)) -> |_, log| {
  log.index_timestamp = t
  result = push(result, log)
}
. = result
```

When `SomeList` was big (30MB), the script took forever to run (more than 30s, meaning quickwit killed my indexing actor, requiring me to restart quickwit-indexer).

So I started to investigate why, and what I've learned is that the VRL interpreter clones all over the place, which is pretty bad for large objects / arrays.

For example:
* Each expression returns a value, even assignment, in case the parent expression wants to use the value.
  * But for expressions inside a block (other than the last), there is no reason to return a value, meaning no reason to clone the entire big array / object.
* `src/compiler/expression/variable.rs` - `Variable::resolve()` clones the variable on lookup of an ident.

At first I thought rewriting everywhere to use `Rc<RefCell<Value>>` but it seemed like too much changes in the source code.

What I ended up with solves only the first example, by running `execute()` instead of `resolve()` which doesn't return a value.

The second example needs more thinking, maybe we can know when a variable is last used (has no references to) after the re-assignment, so we can use `remove_variable()` instead of `variable()`? Can be done in a future PR, as it's completely orthogonal to the optimization in this PR.

Another solution would be to add `map` and `filter`, which won't need to allocate and clone the entire value in each iteration.

The best solution would be to implement a JIT compiler using something like cranelift: https://github.com/vectordotdev/vrl/issues/1095.

DRAFT - still need to prove better performance with benchmarks, but I would still like to hear your opinions.

Seems like most of the time is on the `clone` of 2nd example :(

image

### DIFF
--- a/src/compiler/expression.rs
+++ b/src/compiler/expression.rs
@@ -27,6 +27,7 @@ pub use variable::Variable;
 use crate::value::Value;
 
 use super::state::{TypeInfo, TypeState};
+pub use super::Executed;
 #[allow(clippy::module_name_repetitions)]
 pub use super::ExpressionError;
 pub use super::Resolved;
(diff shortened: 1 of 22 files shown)


tontinton:
Closing as the performance benefits are not that great from my testing.

Also found out `map_values` can work on array and solves my problem:

```py
t = now()
. = map_values(array!(.)) -> |log| {
  log.index_timestamp = t
  log
}
```

tontinton:
Closing as the performance benefits are not that great from my testing.

Also found out `map_values` can work on array and solves my problem:

```py
t = now()
. = map_values(array!(.)) -> |log| {
  log.index_timestamp = t
  log
}
```

pront:
Regardless of the results, thank you for the nice explanation @tontinton.

PR #1332, opened 2025-03-10
chore(build script): sort grok pattern file names

veehaitch:
The `read_grok_patterns` build script function uses `std::fs::read_dir` to read grok patterns from the `src/datadog/grok/patterns` directory and creates the `pattern.rs` file from them.

As mentioned in the [`std::fs::read_dir`](https://doc.rust-lang.org/stable/std/fs/fn.read_dir.html) documentation, "the order in which this iterator returns entries is platform and filesystem dependent", which means that the order of the entries in the resulting `pattern.rs` may be different for the same platform on a different filesystem. Since the patterns end up in the final binary, this poses a reproducibility problem.

To mitigate this problem, this change sorts the result of `std::fs::read_dir` before creating `pattern.rs`.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## Is this a breaking change?

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -31,9 +31,15 @@ fn read_grok_patterns() {
         "#[allow(clippy::needless_raw_string_hashes)]\nstatic PATTERNS: &[(&str, &str)] = &[\n"
             .to_string();
 
-    fs::read_dir(Path::new("src/datadog/grok/patterns"))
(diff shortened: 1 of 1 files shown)


pront (approved):
LGTM

PR #1334, opened 2025-03-11
enhancement(match_datadog_query): add is_phrase flag to equals method

PSeitz:
## Summary

Update the Filter trait's equals signature to include an `is_phrase` boolean flag.

Without that information there is no way to distinguish between phrased and non-phrased queries. They behave differently in datadog on default fields

This is for matching on default fields (typically `message`) in datadog. The matching behavior for phrases is different than non-phrased. In a phrase the tokens need to be in the same order. E.g.
`Hello nice world` => `"Hello world"` no match
`Hello nice world` => `Hello world` matches

### Alternative Options

There is an alternative solution where the tokenizer would emit multiple tokens (only for default fields):
`Hello world` would expand to the equivalent of`_default_:Hello` AND `_default_:World`.
`"Hello world"` would expand to the equivalent of`_default_:"Hello World"`

## Change Type

- [ ] Bug fix
- [X] New feature
- [ ] Performance

## Is this a breaking change?

- [X] Yes
- [ ] No

- [X] No. A maintainer will apply the "no-changelog" label to this PR.

## Checklist

- [X] Our [CONTRIBUTING.md](https://github.com/vectordotdev/vrl/blob/main/CONTRIBUTING.md) is a good starting place.
- [X] For new VRL functions, please also create a sibling PR in Vector to document the new function.

### DIFF
--- a/src/datadog/filter/filter.rs
+++ b/src/datadog/filter/filter.rs
@@ -22,7 +22,12 @@ pub trait Filter<V: Debug + Send + Sync + Clone + 'static>: DynClone {
     /// # Errors
     ///
     /// Will return `Err` if the query contains an invalid path.
-    fn equals(&self, field: Field, to_match: &str) -> Result<Box<dyn Matcher<V>>, PathParseError>;
+    fn equals(
+        &self,
+        field: Field,
+        to_match: &str,
+        is_phrase: bool,
+    ) -> Result<Box<dyn Matcher<V>>, PathParseError>;
 
     /// Determine whether a value starts with a prefix.
     ///
--- a/src/datadog/filter/filter.rs
+++ b/src/datadog/filter/filter.rs
@@ -22,12 +22,14 @@ pub trait Filter<V: Debug + Send + Sync + Clone + 'static>: DynClone {
     /// # Errors
     ///
     /// Will return `Err` if the query contains an invalid path.
-    fn equals(
-        &self,
-        field: Field,
-        to_match: &str,
-        is_phrase: bool,
-    ) -> Result<Box<dyn Matcher<V>>, PathParseError>;
+    fn equals(&self, field: Field, to_match: &str) -> Result<Box<dyn Matcher<V>>, PathParseError>;
+
+    /// Determine whether a field value equals `phrase`.
+    ///
+    /// # Errors
+    ///
+    /// Will return `Err` if the query contains an invalid path.
+    fn phrase(&self, field: Field, phrase: &str) -> Result<Box<dyn Matcher<V>>, PathParseError>;
 
     /// Determine whether a value starts with a prefix.
     ///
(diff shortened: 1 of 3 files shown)


src/datadog/filter/filter.rs
```diff
@@ -22,7 +22,12 @@ pub trait Filter<V: Debug + Send + Sync + Clone + 'static>: DynClone {
     /// # Errors
     ///
     /// Will return `Err` if the query contains an invalid path.
-    fn equals(&self, field: Field, to_match: &str) -> Result<Box<dyn Matcher<V>>, PathParseError>;
+    fn equals(
+        &self,
+        field: Field,
+        to_match: &str,
+        is_phrase: bool,
```
pront, line 29: Which implementation needs this `bool`? 
Also, please avoid passing flags if possible e.g. we can introduce a new `phrase_equals`.

PSeitz: This is for matching on default fields (typically `message`) in datadog. The matching behavior for phrases is different than non-phrased. In a phrase the tokens need to be in the same order. E.g.
`Hello nice world` => `"Hello world"` no match

Btw. there is an alternative solution where the tokenizer would emit multiple tokens.
`Hello world` would expand to the equivalent of`_default_:Hello` AND `_default_:World`.
`"Hello world"` would expand to the equivalent of`_default_:"Hello World"`

PSeitz: updated to introduce `phrase` fn callback

pront:
Thanks @PSeitz, this looks better now. Correct me if I am wrong, but this a breaking change? Since the filter matcher is now stricter.

src/stdlib/match_datadog_query.rs
```diff
@@ -162,6 +162,29 @@ impl Filter<Value> for VrlFilter {
...
+            Field::Default(_) => {
+                let re = word_regex(phrase);
+                Ok(resolve_value(
+                    buf,
+                    Run::boxed(move |value| match value {
+                        Value::Bytes(val) => re.is_match(&String::from_utf8_lossy(val)),
+                        _ => false,
+                    }),
+                ))
+            }
```
bruceg, line 182: How is this different from the condition in `equals` below. It looks word-for-word identical but maybe I missed something. I'd also like to see some unit test cases that demonstrate the different behavior.

20agbekodo:
> Thanks @PSeitz, this looks better now. Correct me if I am wrong, but this a breaking change? Since the filter matcher is now stricter.

I think yes, but nothing disruptive with what's done in datadog (we want to get more in line with the datadog log explorer behaviour)

src/stdlib/match_datadog_query.rs
PSeitz: It is the same, since I didn't know the impact of the change, so currently it's a non-breaking behavior.
Currently the non-phrase behavior is also behaving incorrectly, they both would need to be adjusted to include tokenization (and probably some adaption on the query parser)

I implemented it partially here https://github.com/DataDog/pomsky/pull/79. For full compatibility I'd need to have a closer look on how percolation tokenizes.

Issue #1257, opened 2025-02-04
Unexpected behaviour of the object_from_array VRL fuction


usa4ev:
### A note for the community

A note for the community

    Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
    If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Problem

VRL function **object_from_array** when used with a single parameter requires unnecessary push to that parameter to work properly. See VRL example in respective section.

What I expect:
```
{
  ...
  "foo": "bar",
  "propstructLen": 1,
  ...
}
```

What I get:
```
{
  ...
  "propstructLen": 0,
  ...
}
```

The function is also not fallible depsite the [doc page](https://vector.dev/docs/reference/vrl/functions/#object_from_array).

### Configuration

```text
config.toml:

[sources.demo_logs]
    type = "demo_logs"
    format = "json"
    lines = [ "line1" ]

[transforms.remap_logs]
	inputs = ["demo_logs"]
	type = "remap"
	file = "/Users/username/test/transform_tj.vrl"

[sinks.print_ok]
    type = "console"
    encoding.codec = "json"
    encoding.json.pretty = true
    inputs = [ "remap_logs" ]


transform_tj.vrl:

proparr = [{"key":"foo", "value":"bar"}]

keyvalarr = []
for_each(proparr) -> |_index, item| {

    PropKey = item.key
    PropValue = item.value

    if is_string(PropKey) {
        keyvalarr = push(keyvalarr, [PropKey, PropValue])
    }
}

# keyvalarr = push(keyvalarr, [null, null])  # <-- when this line is commented out or removed, the next line returns an empty object

propstruct = object_from_array(keyvalarr)
.propstructLen = length(propstruct)
. = merge(., propstruct, deep: false)
```

### Version

vector 0.44.0 (aarch64-apple-darwin 3cdc7c3 2025-01-13 21:26:04.735691656)

### Debug Output

```text
TRACE WHEN object_from_array DOES NOT WORK AS EXPECTED:

2025-02-04T14:55:47.266051Z TRACE vector: Beep.
2025-02-04T14:55:47.266050Z TRACE source{component_kind="source" component_id=demo_logs component_type=demo_logs}: vector_common::internal_event::bytes_received: Bytes received. byte_size=0 protocol=none
[...]

PR #1349, opened 2025-03-19
performance(runtime)!: Copy-on-Write value optimization

Zettroke:
## Summary
This request introduces CoW logic into vrl `Value`. It aims to allow passing `Value` between function by reference without the need for copies. It also makes using `resolve_constant` to precompute expressions viable, just making `Object` and `Array` expressions use it yielded 15% more throughput.

This request requires https://github.com/vectordotdev/vector/pull/22693 to work in vector.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Non-functional (chore, refactoring, docs)
- [x] Performance

## Implementation 
All of it is based on `Arc::make_mut` API. To reduce the number of code changes I decided to use custom `DerefMut` which calls to `Arc::make_mut`. While it can potentially lead to excessive copies it's still strictly better than before. But I will understand if you want a more explicit API for this.
I decided to wrap it up with Arc-only `Object` and `Array` variants. This avoids indirection for trivially copyable scalar values, and `Bytes` are already ref-counted. Also, these variants already include an indirection, so overhead is not that noticeable.

## Benchmarks
I used our set of 120+ VRL transforms used for event normalization.
On average this change yielded almost 2x speedup in pure vrl benchmarks.
I attached the full table with results, sadly I'm not allowed to provide sources of our VRL transforms, so I kindly ask you to benchmark on your rule-sets to see speedups in other use cases

| AVG  | MEDIAN |
|-------|-----------|
| 1.92x | 1.49x     |

<details>
<summary>Benchmark results</summary>

[Google sheets](https://docs.google.com/spreadsheets/d/1GZVuejS_24k9QE4bRshKKQoorGOzyIbsdPJXmuBIavk/edit?usp=sharing)

| ID                                   | Baseline (KElem/s) | CoW (KElem/s) | Speedup |
| ------------------------------------ | ------------------ | ------------- | ------- |
| 9d521b11-8120-42d4-9afe-7ee3c6a49f20 | 23.904             | 198.31        | 8.30    |
| 3cf0e90b-5ce1-42f8-b8cb-7e2988ef7b17 | 30.832             | 174.9         | 5.67    |
| 42427cfe-7188-478b-af5b-949e5f7251cc | 19.891             | 103.94        | 5.23    |
| 78441203-a45f-46ac-b4b1-0445a7ec8ce6 | 39.139             | 191.34        | 4.89    |
| 77d02f9c-1422-4544-b890-f5687f5764bd | 26.434             | 121.72        | 4.60    |
[...]

PR #1399, opened 2025-05-08
fix(type system): Fix function closures not correctly updating `TypeState`

Zettroke:
## Summary

Function closures not updating `TypeState` during the compilation, which can result in incorrect types for local variables.

Minimal reproducible example causing runtime error.
Trying to handle this error with coalesce operator causes compilation error.
[playground](https://playground.vrl.dev/?state=eyJwcm9ncmFtIjoidiA9IFwiXCJcblxuZm9yX2VhY2goe1wiYVwiOiAxfSkgLT4gfGtleSwgdmFsdWV8IHtcbiAgICB2ID0gMFxufVxuXG4ucmVzID0gKFwiIVwiICsgdikgIyA%2FPyBcImZhbGxiYWNrXCIiLCJldmVudCI6e30sImlzX2pzb25sIjpmYWxzZSwiZXJyb3IiOm51bGx9)
```python
v = ""

for_each({"a": 1}) -> |key, value| {
    v = 0
}

.res = ("!" + v) # ?? "fallback"
```

This bug was found while using our local fork with [CoW optimization](https://github.com/vectordotdev/vrl/pull/1349) as it heavily relies on `resolve_constant`.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Performance

## Is this a breaking change?

- [x] Yes
- [ ] No

There might be cases where type checks were passed because of this behavior

Dunno :)

## Checklist


pront (approved):
Thank you @Zettroke!

Can you add a VRL test case and a changelog?

Zettroke:
@pront The deed is done

pront:
This looks great @Zettroke. I will fix the failing CI check, it's the first time it ran against a branch from a fork.

src/compiler/expression/function_call.rs
```diff
@@ -501,30 +501,38 @@ impl<'a> Builder<'a> {
...
-                    Some(details) => state.local.insert_variable(ident.clone(), details),
-                    None => {
-                        state.local.remove_variable(ident);
-                    }
-                });
+            for ident in variables {
+                let variable_details = state
+                    .local
+                    .remove_variable(ident)
+                    .expect("Closure variable must be present");
```
pront, line 515: We can avoid `expect`s in this fucntion. See example: https://github.com/vectordotdev/vrl/pull/1399/commits/e2a00093c5161107ca059a5e51010cf397cdf0a5.

Also, do you have any example where this change can break existing behavior? I wonder if we should mark this as a breaking PR.

Zettroke: @pront After some work of adapting our VRL code base to this fix I think I have a better example of broken code.
```
parsed_value = {}

for_each(object!(.value)) -> |key, value| {
    kv = parse_key_value!(value)

    parsed_value = set!(parsed_value, [key], kv)
}

result = ""

for_each(object!(parsed_value)) -> |_asd, value| {
  result = result + value.id + value.level + " " # <- Now this is fallible because type of `value` is unknown
}

.result = result
```

Error:
```
error[E103]: unhandled fallible assignment
   ┌─ :12:12
   │                                                                                                                                                                                                                                                                                                                
12 │   result = result + value.id + value.level + " " # <- Now this is fallible because type of `value` is unknown
   │   -------- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this expression is fallible because at least one argument's type cannot be verified to be valid                                                                                                                                                               
   │   │                                                                                                                                                                                                                                                                                                            
   │   or change this to an infallible assignment:                                                                                                                                                                                                                                                                  
   │   result, err = result + value.id + value.level + " "                                                                                                                                                                                                                                                          
   │                                                                                                                                                                                                                                                                                                                
   = see documentation about error handling at https://errors.vrl.dev/#handling
   = see functions characteristics documentation at https://vrl.dev/expressions/#function-call-characteristics
   = learn more about error code 103 at https://errors.vrl.dev/103
   = see language documentation at https://vrl.dev
   = try your code in the VRL REPL, learn more at https://vrl.dev/examples
```

Zettroke: Also found a trick to keep added error handling compatible with older vrl versions, though it's very ugly 😅
`parsed_value = parse_json!("{}")`
Couldn't think of anything better :)

pront: Thanks, I would be more comfortable if we included this in the next-next release. BTW I fixed `.github/workflows/vector_integration_check.yaml` on main so you can just undo those changes.

pront: What are the before and after types for the `value` (inside the second `for_each` closure) here? And how does `parsed_value = parse_json!("{}")` line changes the type? Just trying to understand if this behavior breakage makes sense before asking all users to change their VRL programs.

pront: [Earlier example](https://playground.vrl.dev/?state=eyJwcm9ncmFtIjoiZWxlbWVudHMgPSBbXVxuXG5mb3JfZWFjaChhcnJheSEoW1wiMVwiLFwiMlwiLFwiM1wiXSkpIC0%2BIHxfaW5kZXgsIHZhbHVlfCB7XG4gICAgZWxlbWVudHMgPSBwdXNoKGVsZW1lbnRzLCB2YWx1ZSlcbiAgICAudDEgPSB0eXBlX2RlZiEodmFsdWUpXG59XG4uZSA9IGVsZW1lbnRzXG4jIGxvZ19zdGF0ZSgpXG5cbmNvbW1hbmRfc3RyaW5nID0gXCJcIlxuZm9yX2VhY2goZWxlbWVudHMpIC0%2BIHxfaW5kZXgsIHZhbHVlfCB7XG4gICAgY29tbWFuZF9zdHJpbmcgPSBjb21tYW5kX3N0cmluZyArIFwiIFwiICsgdmFsdWVcbiAgICAudDIgPSB0eXBlX2RlZiEodmFsdWUpXG59XG5cbi5jID0gY29tbWFuZF9zdHJpbmciLCJldmVudCI6e30sImlzX2pzb25sIjpmYWxzZSwiZXJyb3IiOm51bGx9):

```
	"t1": {
		"bytes": true
	},
	"t2": {
		"never": true
	}
```

Zettroke: `parse_json!("{}")` makes it `any`, so `for_each` can't deduce the type of `(key, value)` pair. Therefore,
 it's type is `never` which forces the error handling.

simplepad:
Fixes #1257

src/compiler/expression/function_call.rs
pront: (Optional) Can you please add a "before and after" in the changelog? It would be very helpful for the next breaking guide and VRL users.

pront (changes requested):
[This is a bug](https://playground.vrl.dev/?state=eyJwcm9ncmFtIjoiZWxlbWVudHMgPSBbXVxuXG5mb3JfZWFjaChhcnJheSEoW1wiMVwiLFwiMlwiLFwiM1wiXSkpIC0%2BIHxfaW5kZXgsIHZhbHVlfCB7XG4gICAgZWxlbWVudHMgPSBwdXNoKGVsZW1lbnRzLCB2YWx1ZSlcbiAgICAudDEgPSB0eXBlX2RlZiEodmFsdWUpXG59XG4uZSA9IGVsZW1lbnRzXG4jIGxvZ19zdGF0ZSgpXG5cbmNvbW1hbmRfc3RyaW5nID0gXCJcIlxuZm9yX2VhY2goZWxlbWVudHMpIC0%2BIHxfaW5kZXgsIHZhbHVlfCB7XG4gICAgY29tbWFuZF9zdHJpbmcgPSBjb21tYW5kX3N0cmluZyArIFwiIFwiICsgdmFsdWVcbiAgICAudDIgPSB0eXBlX2RlZiEodmFsdWUpXG59XG5cbi5jID0gY29tbWFuZF9zdHJpbmciLCJldmVudCI6e30sImlzX2pzb25sIjpmYWxzZSwiZXJyb3IiOm51bGx9):

```json
	"t1": {
		"bytes": true
	},
	"t2": {
		"never": true
	}
```

The type is clearly a `string`. After some internal discussion, I realized that `never` is just the equivalent of Rust's `!` type, which means the compiler can guarantee this value will never exist .

Zettroke:
@pront After a lot of poking around, I found problem in my rebase where I messed up code storing closure argument types for later. 

The example you mentioned now outputs:
```json
{ "t1": { "bytes": true }, "t2": { "bytes": true, "undefined": true } }
```
This is wrong (`t2` should be string), but kinda makes sense, the closure is either called or not. `elements` 0th elements is either a string or undefined.

I think the real problem is in `for_each`. From a type perspective, it should remove `undefined` (and maybe `never` too?) from the value type, because if closure executes, then the value is definitely "defined".

What do you think about this approach? Am I understanding the `undefined` type correctly?

pront:
@codex review

PR #1420, opened 2025-06-08
chore: fix gcc 15 compile

jorgehermo9:
Right now, `vrl` (and therefore, `vector`) fails to compile with GCC 15. This is due to `onig-sys` crate (which is a dependency of `onig` crate, which is a dependency of `grok` crate on which `vrl` depends)

Updating `grok` to `2.1.0` fixes this (https://github.com/mmastrac/grok/commit/9bf239475a735a4b02e13dbba0a48653a227afdf#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542)

More info here https://github.com/rust-onig/rust-onig/issues/195.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -224,7 +224,7 @@ lz4_flex = "0.11.3"
 dns-lookup = { version = "2", optional = true }
 domain = { version = "0.10.4", optional = true, features = ["resolv-sync", "serde"] }
 hostname = { version = "0.4", optional = true }
-grok = { version = "2.1.0", optional = true }
+grok = { version = "2.1", optional = true }
 onig = { version = "6", default-features = false, optional = true }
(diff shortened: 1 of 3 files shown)


pront (approved):

Cargo.toml
```diff
@@ -224,7 +224,7 @@ lz4_flex = "0.11.3"
 dns-lookup = { version = "2", optional = true }
 domain = { version = "0.10.4", optional = true, features = ["resolv-sync", "serde"] }
 hostname = { version = "0.4", optional = true }
-grok = { version = "2", optional = true }
+grok = { version = "2.1.0", optional = true }
```
pront:
```suggestion
grok = { version = "2.1", optional = true }
```

PR #1443, opened 2025-07-04
feat: add `validate_json_schema` function

jlambatl:
## validate_json_schema() function support

This pull request creates a VRL function that validates a payload against a [JSON schema](https://json-schema.org/) definition file. The primary goal of this feature has been to provide a mechanism for users to quickly assert that the schema matches what they expect; this simplifies various error-handling conditions into a single function.

The capability validates that the format (if included in the schema definition) matches and allows the user to determine how to handle non-standard types.

Notes:
* This change has a cache within the function; the goal of this is to improve performance. The validator compiles the schema so that subsequent requests use the schema path as the key. When the next validator is called, it uses an already compiled version if the path matches; otherwise, it creates a new validator. I looked at other implementations across the repository and wasn't sure of the idiomatic way that this should be handled. Other functions (protobuf ones) have this built into the library used, so it was a little guesswork on my part.
* For consistency, the jsonschema crate has been included [without the ability to call out and download schemas](https://docs.rs/jsonschema/latest/jsonschema/index.html#external-references); a schema must be a complete reference, which in some cases may require concatenation or merging by the user.

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Performance

## Is this a breaking change?

- [ ] Yes
- [x] No

## How did you test this PR?

I have created several test cases in the same file as the function that test various features.

## Does this PR include user-facing changes?

## Checklist


changelog.d/1443.feature.md
```diff
@@ -0,0 +1,9 @@
+The addition of `validate_json_schema` makes validation of incoming JSON payloads against JSON Schema definition files possible.
+
+Scenario: `validate_json_schema` with valid payloads returns true
+Scenario: `validate_json_schema` with an invalid payload returns false
+
+The default behaviour is to validate the schema [format](https://json-schema.org/understanding-json-schema/reference/type#format) fields if they exist. A optional configuration parameter `ignore_unknown_formats` is provided to change how custom formats are handled by the validator. `false` (default) considers that unknown format fields are validation errors. Unknown formats
+can be silently ignored by setting this to `true`.
+
+authors: jlambatl
```
pront:
```suggestion
Add `validate_json_schema` function for validating JSON payloads against JSON schema files. A optional configuration parameter `ignore_unknown_formats` is provided to change how custom formats are handled by the validator. Unknown formats can be silently ignored by setting this to `true` and validation continues without failing due to those fields.

authors: jlambatl
```

src/stdlib/validate_json_schema.rs
```diff
@@ -0,0 +1,240 @@
...
+use std::env;
+use std::ffi::OsString;
+use std::path::{Path, PathBuf};
+use std::sync::{Arc, LazyLock, RwLock};
+
+use jsonschema;
+
+// Global cache for compiled schema validators, this allows us to reuse the compiled
+// schema across multiple calls to the function, which is important for performance.
+static SCHEMA_CACHE: LazyLock<RwLock<HashMap<String, Arc<jsonschema::Validator>>>> =
```
pront: One caveat here is that if the underlying schema changes, the process will need to be restarted in order to pick up the new schema. I think this is fine (hashing contents and comparing is probably an overkill here), but we should document it.

src/stdlib/validate_json_schema.rs
```diff
@@ -0,0 +1,240 @@
...
+    // Compile schema
+    let compiled_schema = jsonschema::options()
+        .should_validate_formats(true)
+        .should_ignore_unknown_formats(ignore_unknown_formats)
+        .build(schema_definition)
+        .map_err(|e| format!("Failed to compile schema: {e}"))?;
+
+    let compiled_schema = Arc::new(compiled_schema);
+    cache.insert(schema_path.to_string(), compiled_schema.clone());
+    Ok(compiled_schema.clone())
```
pront: nit: 

```suggestion
    Ok(compiled_schema)
```

src/stdlib/validate_json_schema.rs
```diff
@@ -0,0 +1,240 @@
...
+            .optional("ignore_unknown_formats")
+            .unwrap_or(expr!(false));
+
+        let schema_file_str = schema_file
+            .try_bytes_utf8_lossy()
+            .expect("schema definition file must be a string");
+
+        let os_string: OsString = schema_file_str.into_owned().into();
+        let path_buf = PathBuf::from(os_string);
+        let path = Path::new(&path_buf);
```
pront: Would something like this work? 
```suggestion     
let path = Path::new(schema_file_str.as_ref());
```

src/stdlib/validate_json_schema.rs
```diff
@@ -0,0 +1,240 @@
...
+        ]
+    }
+}
+
+#[derive(Debug, Clone)]
+struct ValidateJsonSchemaFn {
+    value: Box<dyn Expression>,
+    schema_definition: serde_json::Value,
+    ignore_unknown_formats: Box<dyn Expression>,
+    schema_path: String, // Path to the schema file, used for caching
```
pront: In an effort to use `&Path` (vs String, OsString etc), we could use `PathBuf` as a key and pass a `schema_path: &Path`.

src/stdlib/validate_json_schema.rs
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    test_function![
+        validate_json_schema => ValidateJsonSchema;
+
+        valid_with_email_format_json {
+            args: func_args![
+                value: value!("{\"productUser\":\"email@domain.com\"}"),
+                schema_definition: test_data_dir().join("validate_json_schema/schema_with_format_email.json").to_str().unwrap().to_owned(),
+                ignore_unknown_formats: false],
```
pront, line 227: worth adding a test for the `ignore_unknown_formats: true` case

pront (commented):
Thank you @jlambatl! Left a few comments, but nothing major.

src/stdlib/validate_json_schema.rs
jlambatl: @pront, that's a good callout. I did actually think about this and forgot to add it to my notes on the original PR. 

I couldn't find an example of how this has been handled in other functions. I thought that maybe there was an idiomatic way of listening for signals, HUP for reload, which would then flush the cache. Are you aware of any examples of this, or is this a bad idea... suggestions welcome.

src/stdlib/validate_json_schema.rs
jlambatl: I think I've done this correctly based on your feedback, could you have a quick look and verify that what you wrote is what I understood it to be?

pront (approved):
Awesome

pront:
```
For new VRL functions, please also create a sibling PR in Vector to document the new function.
```

Can you followup after this PR is merged? 🙏

jlambatl:
The script is failing the license check because `MIT-0` is not explicitly permitted in `deny.toml`. 

MIT-0 should be compatible with MIT, the primary change is that it doesn't require attribution like MIT does. Here's the license declaration [https://github.com/aws/mit-0](https://github.com/aws/mit-0). 

I am unsure how to proceed on this. Given that it's licensing-related, I don't think I should update the files without first raising this. I am keen to hear your suggestion on what the best next step is.

I think the errors, apart from the one above, should be resolved. 

I have a branch waiting to create a PR in the main vector repository, I'll finalise that tomorrow

jlambatl:
The PR for Vector can be found here: https://github.com/vectordotdev/vector/pull/23359 

I've updated the license check_deny configuration, and that is now passing. looking at why the wasm test/build is not passing.

PR #1447, opened 2025-07-15
enhancement!(lz4): Improve handling of various lz4 implementations

jlambatl:
## Summary

This pull request improves encode_lz4() and decode_lz4() support to handle a variety of different implementations that we have found in production. The original implementation that I contributed prepended the uncompressed data size in bytes to the first four bytes of the compressed payload. This worked for most of the LZ4-encoded payloads that we received; however, it turned out that this pattern is not a standard.

This change introduces a `prepend_size` boolean argument that allows the user to turn on/off expecting the size to be prepended. Because it's non-standard, this defaults to `false`. This means that the change is **breaking**; however, it should increase compatibility.

I have also introduced a `buf_size` argument that sets the buffer size to which the uncompressed data is written. If the user has not set it, then the default is 1MB. 1MB has been chosen because it's larger than many of the common payload sizes (256kb) seen across cloud pub/sub products. If users are familiar with their pipeline, they can adjust it up or down, which may also improve performance on busy systems.

Note, I've used `buf_size` and `prepend_size` as the argument names, I'm unsure how obvious their use case is from the argument names, maybe they need to change? 

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Performance

## Is this a breaking change?

- [x] Yes
- [ ] No

## How did you test this PR?

I have introduced a number of new tests to cover the various differences.

## Does this PR include user-facing changes?

## Checklist

### DIFF
--- a/changelog.d/1447.breaking.md
+++ b/changelog.d/1447.breaking.md
@@ -1,3 +1,5 @@
-`encode_lz4`  no longer prepends the uncompressed size by default, improving compatibility with standard LZ4 tools. A new `prepend_size` flag restores the old behavior if needed. Also, `decode_lz4 now also accepts `prepend_size` and a `buf_size` option (default: 1MB).
+`encode_lz4`  no longer prepends the uncompressed size by default, improving compatibility with standard LZ4 tools. A new `prepend_size` flag restores the old behavior if needed. Also, `decode_lz4` now also accepts `prepend_size` and a `buf_size` option (default: 1MB). 
+
+Existing users of `encode_lz4` and `decode_lz4` will need to update their functions to include the argument `prepend_size: true` to maintain existing compatibility.
 
 authors: jlambatl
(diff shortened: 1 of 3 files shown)


changelog.d/1447.breaking.md
```diff
@@ -0,0 +1,3 @@
+`encode_lz4` used the common method of prepending the uncompressed size to the first four bytes of the compressed payload. This allowed the consumer to decompress the payload, knowing the final size of the uncompressed data, and is fairly performant. Unfortunately, it's a pattern that is not universal and has resulted in payloads that couldn't be decompressed by a number of common libraries. An argument prepend_size has been introduced to turn this behaviour on or off, defaulting to false. decode_lz4 has the same argument, allowing for decoding of lz4 payloads that don't have the uncompressed size prepended. Decode also allows users to specify the buf_size (defaults to 1MB).
```
pront:
```suggestion
`encode_lz4`  no longer prepends the uncompressed size by default, improving compatibility with standard LZ4 tools. A new `prepend_size` flag restores the old behavior if needed. Also, `decode_lz4 now also accepts `prepend_size` and a `buf_size` option (default: 1MB).
```

pront (approved):

changelog.d/1447.breaking.md
pront: Can we also provide a quick example here to inform existing users how to easily upgrade their VRL programs?

jlambatl:
@pront I've updated the changelog fragment to include a statement about compatibility, does that answer your question regarding an example?

Issue #1386, opened 2025-04-23
Add bitwise operators for integers


vruello:
### A note for the community

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

### Use Cases

Logs often contain flags, which are integers where each bit has a special meaning. When processing a flag, it is useful to be able to know whether the n-th bit is set or not. The standard way to do this is to use a bitwise and operator.

### Attempted Solutions

I couldn't find any bitwise operators in the vrl language.

### Proposal

I suggest adding bitwise `and`, `or` and `not` operators for integers in the vrl language.

PR #1504, opened 2025-09-09
feat(operators): Add AND, OR, and NOT bitwise operators

AntoineSauzeau:
## Summary

Add AND (&), OR (^) and NOT (~) bitwise operators. See #1386.

'|' token was already used for the merge operator so I chose another token for the not operator, but it's not ideal since it's usually used for XOR operator.

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Performance

## Is this a breaking change?

- [ ] Yes
- [x] No

## How did you test this PR?

Tests are included.

## Checklist


lib/fuzz/inputs.txt
```diff
@@ -43,4 +43,7 @@ decode_zlib!(decode_base64!("eJwNy4ENwCAIBMCNXIlQ/KqplUSgCdvXAS41qPMHshCB2R1zJlW
 decode_zstd!(decode_base64!("KLUv/QBY/QEAYsQOFKClbQBedqXsb96EWDax/f/F/z+gNU4ZTInaUeAj82KqPFjUzKqhcfDqAIsLvAsnY1bI/N2mHzDixRQA"))
 encode_base16("please encode me")
 encode_base64("please encode me")
-encode_base64(encode_gzip("please encode me"))
\ No newline at end of file
+encode_base64(encode_gzip("please encode me"))
+15 & 20
```
pront, line 47: Nice addition. Did you have a chance to run the fuzzer as well?

pront:
> '|' token was already used for the merge operator so I chose another token for the not operator, but it's not ideal since it's usually used for XOR operator.

Did you mean you chose `^` for the `OR` operator? We want to reserve `^` as the XOR operator. 

I don't see an easy solution around this. Perhaps we could replace `|` with something like `++` and then use `|` for OR but that is a multi-step process. Ideally we would also provide a script for users to update their programs.

lib/fuzz/inputs.txt
asauzeau: Yes I tested it for about ten minutes and didn't experience any crashes. I'll try to run it longer.

asauzeau:
> BTW, I mentioned ++ because it seems like a natural fit, but didn't think too hard. There might be an even better option.

Initially I had in mind `<>` as an alternative, but that might be less intuitive. If there are no plans to add increment and decrement operators, I also think we could opt for `++` for the merge operator.

PR #1531, opened 2025-10-01
feat(stdlib): add new VRL function `basename` to get last component of the path 

titaneric:
## Summary

This PR intends to add a new VRL function called `basename` similar to what Linux command does to get the last component of the given path.

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Performance

## Is this a breaking change?

- [ ] Yes
- [x] No

## How did you test this PR?

```bash
cargo test stdlib --lib
./scripts/check_changelog_fragments.sh
```

## Checklist

### DIFF
--- a/src/stdlib/basename.rs
+++ b/src/stdlib/basename.rs
@@ -1,0 +1,152 @@
+use crate::compiler::prelude::*;
+use std::path::Path;
+
+fn basename(value: &Value) -> Resolved {
+    let path_str_cow = value.try_bytes_utf8_lossy()?;
+    let path_str = path_str_cow.as_ref();
+
+    // strip the trailing slash if it exists
+    let path = if path_str.len() > 1 && path_str.ends_with('/') {
+        Path::new(&path_str[..path_str.len() - 1])
+    } else {
+        Path::new(path_str)
+    };
+
+    let basename = path.file_name().and_then(|s| s.to_str()).map(Value::from);
+    Ok(basename.into())
+}
+
+#[derive(Clone, Copy, Debug)]
+pub struct BaseName;
+
+impl Function for BaseName {
+    fn identifier(&self) -> &'static str {
+        "basename"
+    }
+
+    fn parameters(&self) -> &'static [Parameter] {
+        &[Parameter {
+            keyword: "value",
+            kind: kind::BYTES,
+            required: true,
+        }]
+    }
+
+    fn compile(
+        &self,
+        _state: &state::TypeState,
+        _ctx: &mut FunctionCompileContext,
+        arguments: ArgumentList,
+    ) -> Compiled {
+        let value = arguments.required("value");
+
+        Ok(BaseNameFn { value }.as_expr())
+    }
+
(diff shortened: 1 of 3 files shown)


src/stdlib/basename.rs
```diff
@@ -0,0 +1,152 @@
+use crate::compiler::prelude::*;
+use std::path::Path;
+
+fn basename(value: &Value) -> Resolved {
```
pront, line 4: Can we use https://doc.rust-lang.org/std/path/struct.Path.html#method.file_name here?

titaneric: I have adopted `file_name` in the implemetation, and I remove the necessary trailing slash futher.

pront (approved):
Looks good and sufficiently tested.

titaneric:
@simplepad , providing a additional "platform" sounds a good idea for me. I would dig into it and submit a PR later

simplepad:
@titaneric you would also have to give up on using std::path, and either implement the functions yourself or look into adding a dependency on something like https://github.com/chipsenkbeil/typed-path

pront:
My thoughts, before we rush into an implementation:

* making these platform-agnostic is actually desirable and can avoid bad surprises.

For example we avoid this case:

```
split_path!("C:\\Program Files\\App")
```

Which I believe currently will return:

```
["C:\\Program Files\\App"]
```

* the input can determines the parsing, e.g. `"C:\\Program Files\\App"` is going to be treated as a Windows path. I think `uni_path` will work nicely here.
* we assume correctly formatted input path, paths like `"C:\\Program Files/App"` are left to the implementation and it's up to the user to pre-process/fix such paths (that's where a `to_unix_path` function would come in handy)

simplepad:
@pront Good idea, I just think being able to hint to the function the correct platform to use could improve performance (the function doesn't have to waste cpu cycles guessing) and also resovle any edge cases with relative paths (for example `foo\bar/baz`).

titaneric:
I just give it a try. I think it might impossible to introduce `uni_path` to the VRL

We have multiple compilation error in different functions since the `uni_path` implements [AsRef for Cow](https://docs.rs/uni-path/latest/uni_path/struct.Path.html#impl-AsRef%3CPath%3E-for-Cow%3C'_,+str%3E) and we need to revise multiple of them.
```
error[E0283]: type annotations needed
  --> src/stdlib/parse_duration.rs:13:14
   |
13 |             .get(string.as_ref())
   |              ^^^        ------ type must be known at this point
   |              |
   |              cannot infer type of the type parameter `Q` declared on the method `get`
   |
   = note: multiple `impl`s satisfying `Cow<'_, str>: AsRef<_>` found in the following crates: `alloc`, `uni_path`:
           - impl AsRef<uni_path::Path> for Cow<'_, str>;
           - impl<T> AsRef<T> for Cow<'_, T>
             where T: ToOwned, T: ?Sized;
help: consider specifying the generic argument
   |
13 |             .get::<Q>(string.as_ref())
```

PR #1534, opened 2025-10-06
enhancement(stdlib): Add optional `http_proxy` and `https_proxy` params to `http_request`

5Dev24:
## Summary

Adds optional parameters for setting a proxy for http and https requests to `http_request`.

A future enhancement should use the [global proxy](https://github.com/vectordotdev/vector/blob/cd2471ab3ed81cc55b13eb4f094af735a210b61e/lib/vector-core/src/config/global_options.rs#L103) by default if set. Is there any easy way to pass that information from vector to vrl?

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Performance

## Is this a breaking change?

- [ ] Yes
- [x] No

## How did you test this PR?

Locally testing since I didn't want to add standing up a whole proxy server as party of the test suite.

## Checklist

## References

None

### DIFF
--- a/src/stdlib/http_request.rs
+++ b/src/stdlib/http_request.rs
@@ -10,13 +10,13 @@ use crate::compiler::prelude::*;
 #[cfg(not(target_arch = "wasm32"))]
 mod non_wasm {
     use super::{
-        Context, Expression, FunctionExpression, Resolved, TypeDef, TypeState, Value,
-        VrlValueConvert,
+        Context, Expression, ExpressionError, FunctionExpression, Resolved, TypeDef, TypeState,
+        Value, VrlValueConvert,
     };
     use reqwest_middleware::{
         ClientBuilder, ClientWithMiddleware,
         reqwest::{
-            Client, Method,
+            Client, Method, Proxy,
             header::{HeaderMap, HeaderName, HeaderValue},
         },
     };
@@ -26,21 +26,37 @@ mod non_wasm {
     use tokio::time::Duration;
     use tokio::{runtime, task};
 
-    static CLIENT: LazyLock<ClientWithMiddleware> = LazyLock::new(|| {
+    static STD_CLIENT: LazyLock<ClientWithMiddleware> = LazyLock::new(|| build_client(None));
+
+    fn build_client(proxies: Option<Vec<Proxy>>) -> ClientWithMiddleware {
         let retry_policy = ExponentialBackoff::builder().build_with_max_retries(3);
 
-        let client = Client::builder()
+        let mut client_builder = Client::builder()
             .connect_timeout(Duration::from_secs(5))
-            .timeout(Duration::from_secs(10))
+            .timeout(Duration::from_secs(10));
+
+        if let Some(proxies) = proxies {
+            for proxy in proxies {
+                client_builder = client_builder.proxy(proxy);
+            }
+        }
+
+        let client = client_builder
             .build()
             .expect("Failed to create HTTP client");
 
         ClientBuilder::new(client)
             .with(RetryTransientMiddleware::new_with_policy(retry_policy))
             .build()
-    });
+    }
 
-    async fn http_request(url: &Value, method: &Value, headers: Value, body: &Value) -> Resolved {
+    async fn http_request(
+        client: &ClientWithMiddleware,
+        url: &Value,
+        method: &Value,
+        headers: Value,
+        body: &Value,
+    ) -> Resolved {
         let url = url.try_bytes_utf8_lossy()?;
         let method = method.try_bytes_utf8_lossy()?.to_uppercase();
         let headers = headers.try_object()?;
@@ -61,7 +77,7 @@ mod non_wasm {
             header_map.insert(key, val);
         }
 
-        let response = CLIENT
+        let response = client
             .request(method, url.as_ref())
             .headers(header_map)
             .body(body.as_bytes().to_owned())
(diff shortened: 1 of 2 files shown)


src/stdlib/http_request.rs
```diff
@@ -61,7 +77,7 @@ mod non_wasm {
             header_map.insert(key, val);
         }
 
-        let response = CLIENT
+        let response = client
             .request(method, url.as_ref())
```
pront, line 65: Pre-existing issue: Redundant `to_uppercase` call.

src/stdlib/http_request.rs
```diff
@@ -73,37 +89,89 @@ mod non_wasm {
...
+    #[allow(clippy::similar_names)]
+    fn make_proxies(
+        http_proxy: Option<Value>,
+        https_proxy: Option<Value>,
+    ) -> Result<Option<Vec<Proxy>>, ExpressionError> {
+        let mut proxies = Vec::new();
+
+        if let Some(http_proxy) = http_proxy {
+            proxies.push(
+                Proxy::http(http_proxy.try_bytes_utf8_lossy()?.into_owned())
```
pront: `into_owned` looks redundant here

src/stdlib/http_request.rs
```diff
@@ -73,37 +89,89 @@ mod non_wasm {
...
+                    .map_err(|e| format!("Invalid proxy: {e}"))?,
+            );
+        }
+
+        if let Some(https_proxy) = https_proxy {
+            proxies.push(
+                Proxy::https(https_proxy.try_bytes_utf8_lossy()?.into_owned())
```
pront: `into_owned` looks redundant here

src/stdlib/http_request.rs
```diff
@@ -73,37 +89,89 @@ mod non_wasm {
...
+                .map(|http_proxy| http_proxy.resolve(ctx))
+                .transpose()?;
+            let https_proxy = self
+                .https_proxy
+                .as_ref()
+                .map(|https_proxy| https_proxy.resolve(ctx))
+                .transpose()?;
+
+            let client = if let Some(proxies) = make_proxies(http_proxy, https_proxy)? {
+                &build_client(Some(proxies))
```
pront: We are building a client on each request here. It should be possible to detect literals in `compile`, call `resolve_constant` and create the client there.

pront (commented):
A few problems but overall looks good

src/stdlib/http_request.rs
5Dev24: `Proxy::http` takes any `dyn IntoProxy` which is implemented for any `dyn IntoUrl`. `IntoUrl` is only implemented for `url::Url`, `String`, `&str`, and `&String`. `into_owned` was to take the `Cow<str>` as a `String`, but I can substitute for `as_ref` to avoid the copy.

src/stdlib/http_request.rs
5Dev24: Ditto above.

src/stdlib/http_request.rs
pront: `as_ref` sounds better

src/stdlib/http_request.rs
5Dev24: Removed

src/stdlib/http_request.rs
5Dev24: Added in logic to attempt to resolve the values as constants.

src/stdlib/http_request.rs
```diff
@@ -181,26 +341,46 @@ impl Function for HttpRequest {
...
         arguments: ArgumentList,
     ) -> Compiled {
         let url = arguments.required("url");
         let method = arguments.optional("method").unwrap_or_else(|| expr!("get"));
         let headers = arguments.optional("headers").unwrap_or_else(|| expr!({}));
         let body = arguments.optional("body").unwrap_or_else(|| expr!(""));
+        let http_proxy = arguments.optional("http_proxy");
+        let https_proxy = arguments.optional("https_proxy");
+
+        // Result::map_err wouldn't properly convert the error for ? sugar
```
pront: I am not following this comment, do you have an example?

src/stdlib/http_request.rs
```diff
@@ -73,37 +89,181 @@ mod non_wasm {
...
+                        .map(|http_proxy| http_proxy.resolve(ctx))
+                        .transpose()?;
+
+                    let https_proxy = https_proxy
+                        .as_ref()
+                        .map(|https_proxy| https_proxy.resolve(ctx))
+                        .transpose()?;
+
+                    if let Some(proxies) = make_proxies(http_proxy, https_proxy)? {
+                        Ok(RefOrOwnedClient::Owned(build_client(Some(proxies))))
```
pront: It's possible to build a client per request, is this something you require? Or can we remove this from here?

pront (commented):
forgot to submit these comments

src/stdlib/http_request.rs
5Dev24: That client being built happens in the case where the proxies aren't constant. We wouldn't be able to make the client ahead of time since we don't know what proxy to use yet.

If we resolve it once and save the value (inside some `Arc<Mutex<T>>` since we don't have a mut ref to the `HttpRequestFn`), we risk saving the wrong proxy for a future request.

Example of time we can't cache client (you can also imagine the case where the proxy is extracted from the event itself):

```vrl
log = parse_json!(.message)

proxy = if log.type == "value1" {
  "proxy-server-a.com"
} else if log.type == "value2" {
  "proxy-server-b.com"
} else {
  "proxy-server-fallback.com"
};

.message = http_proxy(..., http_proxy: proxy)
```

We could recommend instead doing:

```vrl
log = parse_json!(.message)

.message = if log.type == "value1" {
  http_proxy(..., http_proxy: "proxy-server-a.com")
} else if log.type == "value2" {
  http_proxy(..., http_proxy: "proxy-server-b.com")
} else {
  http_proxy(..., http_proxy: "proxy-server-fallback.com")
};
```

src/stdlib/http_request.rs
5Dev24: Disregard, it needed an explicit cast. I could've sworn I tried it previously and still got compiler errors.

src/stdlib/http_request.rs
```diff
@@ -73,37 +89,181 @@ mod non_wasm {
...
+        }
+
+        if proxies.is_empty() {
+            Ok(None)
+        } else {
+            Ok(Some(proxies))
+        }
+    }
+
+    // Used to avoid clones
```
pront: `ClientWithMiddleware` only holds `Arc`s internally. I will push a commit to delete `RefOrOwnedClient`.

pront (approved):
Pushed commit 7a5cdad86cf1a72202743a5ba6662275d6bfbe25.
@5Dev24 this looks good to merge now. 
Feel free to follow up if something was missed.

src/stdlib/http_request.rs
5Dev24: Sounds good.

PR #1542, opened 2025-10-20
fix(datadog search): Fix improper handling of conjunctions by the datadog query parser

gwenaskell:
## Summary

The Datadog query parser incorrectly interprets queries mixing AND and OR statements.

For example:

- it interprets A OR NOT B as A AND NOT B
- it interprets A AND B OR C as A AND (B OR C)
- it interprets A OR B AND C as B AND C
- it interprets A OR B C as C
- it interprets A B OR C, as well as A AND B OR C as A

This PR fixes this behavior, while simplifying the parser logic.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Performance

## Is this a breaking change?

- [x] Yes
- [ ] No

## How did you test this PR?

Added and updated unit tests

## Checklist

### DIFF
--- a/src/datadog/search/parser.rs
+++ b/src/datadog/search/parser.rs
@@ -158,21 +158,15 @@ mod tests {
         let cases = ["foo bar baz AND qux quux quuz"];
         for query in cases.iter() {
             let res = parse(query);
-            if let QueryNode::Boolean {
-                oper: BooleanType::And,
-                ref nodes,
-            } = res
-            {
-                assert!(
+            assert!(
+                matches!(res, QueryNode::Boolean { oper: BooleanType::And, ref nodes } if
                     matches!(nodes[0], QueryNode::AttributeTerm { ref attr, ref value } if attr == "_default_" && value == "foo bar")
                         && matches!(nodes[1], QueryNode::AttributeTerm { ref attr, ref value } if attr == "_default_" && value == "baz")
                         && matches!(nodes[2], QueryNode::AttributeTerm { ref attr, ref value } if attr == "_default_" && value == "qux")
-                        && matches!(nodes[3], QueryNode::AttributeTerm { ref attr, ref value } if attr == "_default_" && value == "quux quuz"),
-                    "Unable to properly parse '{query:?}' - got {res:?}"
-                );
-            } else {
-                panic!("Unable to properly parse '{query:?}' - got {res:?}")
-            }
+                        && matches!(nodes[3], QueryNode::AttributeTerm { ref attr, ref value } if attr == "_default_" && value == "quux quuz")
+                ),
+                "Unable to properly parse '{query:?}' - got {res:?}"
+            );
         }
     }
 
@@ -477,20 +471,14 @@ mod tests {
         let cases = ["foo:bar baz:qux quux:quuz"];
         for query in cases.iter() {
             let res = parse(query);
-            if let QueryNode::Boolean {
-                oper: BooleanType::And,
-                ref nodes,
-            } = res
-            {
-                assert!(
+            assert!(
+                matches!(res, QueryNode::Boolean { oper: BooleanType::And, ref nodes } if
                     matches!(nodes[0], QueryNode::AttributeTerm { ref attr, ref value } if attr == "foo" && value == "bar")
                         && matches!(nodes[1], QueryNode::AttributeTerm { ref attr, ref value } if attr == "baz" && value == "qux")
-                        && matches!(nodes[2], QueryNode::AttributeTerm { ref attr, ref value } if attr == "quux" && value == "quuz"),
-                    "Unable to properly parse '{query:?}' - got {res:?}"
-                );
-            } else {
-                panic!("Unable to properly parse '{query:?}' - got {res:?}")
-            }
+                        && matches!(nodes[2], QueryNode::AttributeTerm { ref attr, ref value } if attr == "quux" && value == "quuz")
+                ),
+                "Unable to properly parse '{query:?}' - got {res:?}"
+            );
         }
     }
 
@@ -504,20 +492,14 @@ mod tests {
         ];
         for query in cases.iter() {
             let res = parse(query);
-            if let QueryNode::Boolean {
-                oper: BooleanType::And,
-                ref nodes,
-            } = res
-            {
-                assert!(
+            assert!(
+                matches!(res, QueryNode::Boolean { oper: BooleanType::And, ref nodes } if
                     matches!(nodes[0], QueryNode::NegatedNode { ref node } if matches!(**node, QueryNode::AttributeTerm {ref attr, ref value } if attr == "foo" && value == "bar"))
                         && matches!(nodes[1], QueryNode::AttributeTerm { ref attr, ref value } if attr == "baz" && value == "qux")
-                        && matches!(nodes[2], QueryNode::NegatedNode { ref node } if matches!(**node, QueryNode::AttributeTerm {ref attr, ref value } if attr == "quux" && value == "quuz")),
-                    "Unable to properly parse '{query:?}' - got {res:?}"
-                );
-            } else {
-                panic!("Unable to properly parse '{query:?}' - got {res:?}")
-            }
+                        && matches!(nodes[2], QueryNode::NegatedNode { ref node } if matches!(**node, QueryNode::AttributeTerm {ref attr, ref value } if attr == "quux" && value == "quuz"))
+                ),
+                "Unable to properly parse '{query:?}' - got {res:?}"
+            );
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/datadog/search/grammar.rs
```diff
@@ -21,6 +21,19 @@ const MISSING_FIELD: &str = "_missing_";
 /// via a Visitor pattern and walking our way through the syntax tree.
 pub struct QueryVisitor;
 
+/// Group a list of nodes into a single node, using the given conjunction.
+/// If the group has only one node, return it as-is.
+fn group_nodes(mut node_group: Vec<QueryNode>, conjunction: BooleanType) -> QueryNode {
```
bruceg: nit: Would this be better expressed as a method on `QueryNode` i.e. something like `QueryNode::new_boolean(…) -> Self`?

src/datadog/search/grammar.rs
```diff
@@ -33,32 +46,28 @@ impl QueryVisitor {
...
-                            // a MUST, we make the previous clause a SHOULD and our
-                            // new clause will also be a SHOULD
-                            let lastitem = clauses.last_mut().unwrap();
-                            if let LuceneOccur::Must = lastitem.occur {
-                                lastitem.occur = LuceneOccur::Should;
-                            };
-                            modifier.get_or_insert(LuceneOccur::Should);
+                            // close the current and_group and create a new one
+                            and_groups.push(group_nodes(and_group, BooleanType::And));
+                            and_group = Vec::new();
```
bruceg, line 57: nit:
```suggestion
                            and_group.clear();
```

src/datadog/search/grammar.rs
```diff
@@ -33,32 +46,28 @@ impl QueryVisitor {
...
-                            // If our conjunction is OR and the previous clause was
-                            // a MUST, we make the previous clause a SHOULD and our
-                            // new clause will also be a SHOULD
-                            let lastitem = clauses.last_mut().unwrap();
-                            if let LuceneOccur::Must = lastitem.occur {
-                                lastitem.occur = LuceneOccur::Should;
-                            };
-                            modifier.get_or_insert(LuceneOccur::Should);
+                            // close the current and_group and create a new one
+                            and_groups.push(group_nodes(and_group, BooleanType::And));
```
bruceg: Should this and below maybe only push if `!and_group.is_empty()` or is that unreachable?

src/datadog/search/grammar.rs
```diff
@@ -69,77 +78,42 @@ impl QueryVisitor {
                     match inner.as_rule() {
                         Rule::PLUS => (),
                         Rule::NOT => {
-                            modifier = Some(LuceneOccur::MustNot);
+                            is_not = true;
```
bruceg, line 68: Does this need to toggle? i.e. `is_not = !is_not;`

src/datadog/search/grammar.rs
```diff
@@ -33,32 +46,28 @@ impl QueryVisitor {
...
+        // Finally, we will combine all the and_groups with OR.
+
+        let mut and_groups: Vec<QueryNode> = Vec::new();
+
+        let mut and_group: Vec<QueryNode> = Vec::new();
+
         for node in contents {
-            let clause: Option<LuceneClause> = match node.as_rule() {
+            let query_node: Option<QueryNode> = match node.as_rule() {
                 Rule::multiterm => Some(Self::visit_multiterm(node, default_field)),
```
bruceg: Do we also need the same "and" logic handling with the implicit "and" joins between the parts of a multiterm?

src/datadog/search/parser.rs
```diff
@@ -591,4 +602,86 @@ mod tests {
...
+                ref nodes,
+            } = res
+            {
+                assert!(
+                    matches!(nodes[0], QueryNode::Boolean { oper: BooleanType::Or, ref nodes } if
+                        matches!(nodes[0], QueryNode::AttributeTerm { ref attr, ref value } if attr == "foo" && value == "bar") &&
+                        matches!(nodes[1], QueryNode::AttributeTerm { ref attr, ref value } if attr == "baz" && value == "qux")
+                    ) && matches!(nodes[1], QueryNode::AttributeTerm {ref attr, ref value } if attr == "quux" && value == "quuz"),
+                    "Unable to properly parse '{query:?}' - got {res:?}"
+                );
```
bruceg: Could this all be joined into one large `assert!(matches!` (with sub-matches)?

src/datadog/search/grammar.rs
gwenaskell: I tried that first but since and_group is moved into group_nodes, I could not reuse it

src/datadog/search/grammar.rs
gwenaskell: Like, a query containing "NOT NOT"? I just checked, the lexer does not allow it (it fails on "NOT -foo" and interprets NOT NOT foo as NOT "NOT" foo)

src/datadog/search/grammar.rs
gwenaskell: I think this is unreachable, because a query needs to contain some clause before an OR statement. The query OR foo is interpreted as "OR" foo by the lexer

src/datadog/search/parser.rs
gwenaskell: I'm not sure how to achieve this 🤔

gwenaskell:
> It does have me wondering if it would need to be something recursive, however, to fit all possible cases of nested parentheses. Could you comment on that?

Parentheses are interpreted by the parser as the beginning of a sub-clause. The parser is already recursive (visit_query -> visit_clause -> visit_query)

src/datadog/search/parser.rs
bruceg: Something like this:
```suggestion
            assert!(matches!(res, QueryNode::Boolean { oper: BooleanType::And, ref nodes }
               if matches!(nodes[0], QueryNode::Boolean { oper: BooleanType::Or, ref nodes } if
                        matches!(nodes[0], QueryNode::AttributeTerm { ref attr, ref value } if attr == "foo" && value == "bar") &&
                        matches!(nodes[1], QueryNode::AttributeTerm { ref attr, ref value } if attr == "baz" && value == "qux")
                    ) && matches!(nodes[1], QueryNode::AttributeTerm {ref attr, ref value } if attr == "quux" && value == "quuz")
                )
            );
```
Not that that's necessarily any easier to read, with all the nested `matches!` clauses.

src/datadog/search/grammar.rs
bruceg: That still makes me nervous, given the prevalence of double-negation in other such expression parsers, but it's good to know it's at least not wrong.

src/datadog/search/grammar.rs
```diff
@@ -69,77 +65,42 @@ impl QueryVisitor {
...
-                        });
-                        has_must_not = true;
-                    }
-                    LuceneOccur::Should => {
-                        or_builder.add_node(node);
-                        has_should = true;
-                    }
+            if let Some(mut n) = query_node {
+                if is_not {
+                    is_not = false;
```
bruceg, line 80: nit: If I'm reading right, this reset is superfluous as the value of `is_not` is not used later in the function.

src/datadog/search/node.rs
```diff
@@ -343,6 +304,19 @@ impl QueryNode {
             format!("{attr}:")
         }
     }
+
+    /// Group a list of nodes into a single node, using the given conjunction.
+    /// If the group has only one node, return a clone of that node.
+    pub fn new_boolean(conjunction: BooleanType, nodes: Vec<QueryNode>) -> QueryNode {
+        if nodes.len() == 1 {
+            return nodes.first().unwrap().clone();
```
bruceg: nit: I think the `clone` can be avoided with either `remove` or `into_iter`, with the latter likely being more performant. In either case, I think we'd prefer an `expect` instead of an `unwrap` just so the code documents why it is safe.
```suggestion
            return nodes.into_iter().next().expect("Known to have length 1");
```

src/datadog/search/grammar.rs
gwenaskell: It can be used multiple times: for example, `-foo -baz`

gwenaskell: The reset is actually useful for that situation: `-foo bar`. Otherwise the query would be interpreted `-foo -bar`.

src/datadog/search/parser.rs
gwenaskell: Oh, I did not understand you were talking about the initial if let. Done.

src/datadog/search/grammar.rs
bruceg: Oh, of course, I was misreading the level of indentation as being outside of the loop. That suggests some possible refactoring for readability, but you are right it is needed.

PR #1554, opened 2025-11-06
chore(stdlib): create macro to keep exports and stdlib in sync

thomasqueirozb:
## Summary

This macro prevents functions from being exported but not included in the stdlib. Should make maintenence better and prevent issues like #1553 from happening

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## Is this a breaking change?

- [ ] Yes
- [x] No

## How did you test this PR?

I verified # of imports before (200) and lines inside the macro (199). It is 199 since `pub use get_timezone_name::get_name_for_timezone` isn't and shouldn't be included inside `all()`.

## Checklist

### DIFF
--- a/changelog.d/1553.fix.md
+++ b/changelog.d/1553.fix.md
@@ -1,0 +1,4 @@
+Fixed an issue where `split_path`, `basename`, `dirname` had not been added to VRL's standard
+library and, therefore, appeared to be missing and were inaccessible in the `0.28` release.
+
+authors: thomasqueirozb
--- a/src/stdlib/split_path.rs
+++ b/src/stdlib/split_path.rs
@@ -42,22 +42,22 @@ impl Function for SplitPath {
         &[
             Example {
                 title: "Split path with trailing slash",
-                source: r#"split_path!("/home/user/")"#,
+                source: r#"split_path("/home/user/")"#,
                 result: Ok(r#"["/", "home", "user"]"#),
             },
             Example {
                 title: "Split path from file path",
-                source: r#"split_path!("/home/user")"#,
+                source: r#"split_path("/home/user")"#,
                 result: Ok(r#"["/", "home", "user"]"#),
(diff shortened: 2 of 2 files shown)


src/stdlib/mod.rs
```diff
@@ -210,412 +210,233 @@ cfg_if::cfg_if! {
...
+            zip::Zip,
+            self::array::Array,
+            self::md5::Md5,
+            self::seahash::Seahash,
+            self::sha1::Sha1,
+            self::xxhash::Xxhash,
+            self::crc::Crc,
+        }
+
         pub use get_timezone_name::get_name_for_timezone;
```
pront, line 440: Why is this `use` special?

src/stdlib/mod.rs
```diff
@@ -210,412 +210,233 @@ cfg_if::cfg_if! {
...
-        pub use format_number::FormatNumber;
-        pub use format_timestamp::FormatTimestamp;
-        pub use from_unix_timestamp::FromUnixTimestamp;
-        pub use self::community_id::CommunityID;
-        pub use get::Get;
-        pub use get_env_var::GetEnvVar;
-        pub use get_hostname::GetHostname;
-        pub use get_timezone_name::GetTimezoneName;
+        // Macro to keep pub use and all() function in sync
+        macro_rules! stdlib_functions {
```
pront, line 214: Did you explore any alternatives that do not need a macro?

thomasqueirozb: I didn't but I don't see how we'd do this without a macro. Open to ideas

src/stdlib/mod.rs
thomasqueirozb: It's the inner function of `GetTimezoneName`, this export is only here to be used in bench code (see `benches/stdlib.rs`)

PR #1579, opened 2025-12-10
fix: properly emit lexer parse errors

thomasqueirozb:
## Summary

Any lexer error was being emitted as a generic `ParseError`.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Performance

## Is this a breaking change?

- [ ] Yes
- [x] No

## How did you test this PR?

```sh
echo '"\\e"' | cargo run -p vrl-cli
```

Before:
```
error[E202]: syntax error
  ┌─ :1:1
  │
1 │ "\e"
  │ ^^^^ unexpected error: invalid escape character: \e
  │
```

After:
```
error[E209]: invalid escape character: \e
  ┌─ :1:3
  │
1 │ "\e"
  │   ^ invalid escape character: e
  │
```

You can also verify that nested lexers work by running

```sh
echo 'test("\\e")' | cargo run -p vrl-cli
```

and verifying the output looks like

```
error[E209]: invalid escape character: \e
  ┌─ :1:8
  │
1 │ test("\e")
  │        ^ invalid escape character: e
  │
  = see language documentation at https://vrl.dev
  = try your code in the VRL REPL, learn more at https://vrl.dev/examples
```

### DIFF
--- a/changelog.d/1579.fix.md
+++ b/changelog.d/1579.fix.md
@@ -2,4 +2,34 @@ Fixed a bug where lexer parse errors would emit a generic span with 202 error co
 proper error. Also fixed error positions from nested lexers (e.g., string literals inside function
 arguments) to correctly point to the actual location in the source.
 
+Before (generic E202 syntax error):
+
+```
+$ string("\a")
+
+error[E202]: syntax error
+  ┌─ :1:1
+  │
+1 │ string("\a")
+  │ ^^^^^^^^^^^^ unexpected error: invalid escape character: \a
+  │
+  = see language documentation at https://vrl.dev
+  = try your code in the VRL REPL, learn more at https://vrl.dev/examples
+```
+
+After (correct E209 invalid escape character):
+
+```
+$ string("\a")
+
+error[E209]: invalid escape character: \a
+  ┌─ :1:10
+  │
+1 │ string("\a")
+  │          ^ invalid escape character: a
+  │
+  = see language documentation at https://vrl.dev
+  = try your code in the VRL REPL, learn more at https://vrl.dev/examples
+```
+
 authors: thomasqueirozb


lib/tests/tests/issues/6571_invalid_escape.vrl
```diff
@@ -1,13 +1,11 @@
...
-# 1 │ ╭
-# 2 │ │ .field = "foo\/bar"
-# 3 │ │
-#   │ ╰^ unexpected error: invalid escape character: \/
-#   │
+# result: error[E209]: invalid escape character: \/
+# ┌─ :2:15
+# │
+# 2 │ .field = "foo\/bar"
+# │               ^ invalid escape character: /
```
jszwedko: This seems to be pointing to one or two characters earlier than the invalid one (depending on if we consider `\` part of the "token"). Is that right?

thomasqueirozb: This is even worse in other tests. It exposed a pre-existing bug. I'll hold off on merging this since now at least the bug is hidden and this just "exposes" it to users

bfung:
Doing some local testing, the `main` branch would "underscore" the whole VRL line:

Screenshot 2025-12-10 at 3 44 54 PM


The change here makes the error surface to the character, however, it seems to be pointing at the wrong character and may cause more confusion if we merge as is:

Screenshot 2025-12-10 at 3 46 03 PM

Headed in right direction - should have this point to the right character(s) before calling it a completed fix.

lib/tests/tests/diagnostics/invalid_escape_char.vrl
```diff
@@ -1,12 +1,9 @@
...
-#   │
-# 1 │ ╭
-# 2 │ │ . |= parse_grok!("1.2.3.4 - - [23/Mar/2021:06:46:35 +0000]", "%{IPORHOST:remote_ip} %{USER:ident} %{USER:user_name} \[%{HTTPDATE:timestamp}\]"
-# 3 │ │
-#   │ ╰^ unexpected error: invalid escape character: \[
+# error[E209]: invalid escape character: \[
+# ┌─ :2:55
+# │
+# 2 │ . |= parse_grok!("1.2.3.4 - - [23/Mar/2021:06:46:35 +0000]", "%{IPORHOST:remote_ip} %{USER:ident} %{USER:user_name} \[%{HTTPDATE:timestamp}\]"
+# │                                                       ^ invalid escape character: [
```
thomasqueirozb: This is incorrect, the `^` is in the wrong place. Should be addressed before this one is merged so that users don't see this. Right now the error is technically incorrect (202 instead of 209) and highlights the whole line but at least it isn't misleading.

lib/tests/tests/issues/6571_invalid_escape.vrl
thomasqueirozb: Update: this one is actually correct. The whitespace is messed up but running it locally shows the following:

```
error[E209]: invalid escape character: \/
  ┌─ :1:15
  │
1 │ .field = "foo\/bar"
  │               ^ invalid escape character: /
  │
  = see language documentation at https://vrl.dev
  = try your code in the VRL REPL, learn more at https://vrl.dev/examples
```
I made #1580 so that whitespace is no longer messed up like this in tests.

However, `parse_grok` is still incorrect.

changelog.d/1579.fix.md
```diff
@@ -0,0 +1,5 @@
+Fixed a bug where lexer parse errors would emit a generic span with 202 error code instead of the
```
pront, line 1: Would be nice to include an example here

thomasqueirozb: done

pront (approved):
Nice improvement for users!

PR #1602, opened 2026-01-09
feat: add `parse_yaml` function

juchem:
Adds `parse_yaml` function as per discussion in https://github.com/vectordotdev/vrl/discussions/1601.

- New feature
- No breaking change

Sibling doc PR: https://github.com/vectordotdev/vector/pull/24479

Tested with:
```
./scripts/checks.sh
```


src/stdlib/parse_yaml.rs
```diff
@@ -0,0 +1,168 @@
+//use serde_yaml::{
+//    //Error,
+//    //value::{Value as YamlValue},
+//};
+
```
thomasqueirozb: We can remove this


```suggestion
```

src/stdlib/parse_yaml.rs
```diff
@@ -0,0 +1,168 @@
...
+    fn parameters(&self) -> &'static [Parameter] {
+        &[Parameter {
+            keyword: "value",
+            kind: kind::BYTES,
+            required: true,
+        }]
+    }
+
+    fn examples(&self) -> &'static [Example] {
```
thomasqueirozb, line 66: These examples are fine but they're missing an actual yaml like you'd expect to see. Something like what you did in the `complex_yaml` test would be nice to have here

Cargo.toml
```diff
@@ -184,6 +184,7 @@ rust_decimal = { version = "1", optional = true, default-features = false }
 seahash = { version = "4", optional = true }
 serde = { version = "1", features = ["derive"], optional = true }
 serde_json = { version = "1", default-features = false, optional = true, features = ["std", "raw_value"] }
+serde_yaml = { version = "0.9.34" }
```
thomasqueirozb: This library is now unmaintained. I think we should use either serde_yaml_ng or serde_norway instead. I haven't looked too deeply into them to know which one is better maintained

thomasqueirozb (commented):
Thanks for this!

src/stdlib/parse_yaml.rs
juchem: I struggled with a raw string inside another since rust is not my primary programming language. What's the best way to do it without the compiler complaining about the extra hashes?

thomasqueirozb: You can do something like this:

```rs
r#"
this_is_a:
  - multiline_string # with comments!
  - "and it even works with quotes"
"#
```

src/stdlib/parse_yaml.rs
juchem: removed

src/stdlib/parse_yaml.rs
juchem: nvm, the issue was something else. added a better example

Cargo.toml
juchem: `serde_yaml_ng` looks more active, switched to latest version of that one

changelog.d/1602.feat.md
```diff
@@ -0,0 +1,3 @@
+Add function `parse_yaml`.
```
thomasqueirozb:
```suggestion
Added a new `parse_yaml` function. This function parses yaml according to the [YAML 1.1 spec](https://yaml.org/spec/1.1/).
```

File also needs to be renamed to 1602.feature.md

src/stdlib/parse_yaml.rs
```diff
@@ -0,0 +1,191 @@
...
+
+    fn usage(&self) -> &'static str {
+        indoc! {"
+            Parses the provided `value` as YAML.
+
+            Only YAML types are returned. If you need to convert a `string` into a `timestamp`,
+            consider the `parse_timestamp` function.
+        "}
+    }
+
```
thomasqueirozb, line 30: There is an ongoing effort in order to make it so the sibling docs PR is no longer needed. This would make it so an implementation of `internal_failure_reasons`, `notices` and `return_type` is needed here (not all of those methods are available in main right now). I can do the implementation once those are merged.

I will hold off on merging this one until that effort is complete which should be before next release. If that's not the case then I'll merge this one and the docs PR too.

thomasqueirozb (commented):
Thanks. There are a few failing checks. I'll circle back to this PR once checks are passing and those other changes I mentioned are fully implemented.

juchem:
It looks like I rebased in the middle of a refactoring.

Here are the tests I gathered from CI jobs. Made some more fixes and pushed them, all of these are passing now on my container:
```
cargo install dd-rust-license-tool \
  && cargo update --workspace --locked \
  && cargo doc --no-deps --workspace \
  && cargo check --no-default-features \
  && cargo check --no-default-features --features __mock_return_values_for_tests \
  && cargo check --no-default-features --features arbitrary \
  && cargo check --no-default-features --features cli \
  && ./scripts/format_check.sh \
  && ./scripts/check_deny.sh \
  && ./scripts/check_wasm32.sh \
  && ./scripts/vrl_tests.sh \
  && ./scripts/tests.sh \
  && ./scripts/clippy.sh \
  && ./scripts/check_msrv.sh \
  && ./scripts/check_licenses.sh
```

thomasqueirozb (approved):

thomasqueirozb:
Docs already approved the Vector PR [here](https://github.com/vectordotdev/vector/pull/24479). No need to review. Merging this in

PR #1610, opened 2026-01-21
feat!(stdlib): add required description field to Parameter

thomasqueirozb:
## Summary

Adds a required `description` field to the `Parameter` struct to document what each function parameter does. This is a breaking change as all `Parameter` instantiations must now include a description.

Part of the automatic documentation generation RFC (https://github.com/vectordotdev/vrl/issues/280)

**Review Note:** This PR is best reviewed commit-by-commit. Some commits contain automated changes and can be skipped during review.

### Commits Overview

1. **718dec5** - Adds the new `description` field to the `Parameter` struct in `src/compiler/function.rs`

2. **d75af2b** - Introduces `update_parameter_descriptions.py` which automates fetching parameter descriptions from Vector's `docs.json` file. This file (basically) generates the next commit.

3. **a1d2ec1** - Automatically generated descriptions. This commit can be skipped. I added all descriptions which matched the cue files but didn't add the TODOs

4. **b74b022** - Create new descriptions for undocumented parameters.

5. **7143948** - Manually add descriptions for parameters that were documented but had a name missmatch (e.g. `ip` vs `value`, `path` vs `value`)

6. **3a5ec83** - Removes the temporary python script

7. **21589cf** and later - formatting changes only

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Non-functional (chore, refactoring, docs)
- [ ] Performance

## Is this a breaking change?

- [x] Yes
- [ ] No

## How did you test this PR?

- Verified code compiles with `cargo check`
- Existing tests pass
- Script successfully populates descriptions from docs.json
- Manual verification of parameter descriptions in sample files

## Does this PR include user facing changes?

- [x] Yes. Please add a changelog fragment based on
  our [guidelines](https://github.com/vectordotdev/vector/blob/master/changelog.d/README.md).
- [ ] No. A maintainer will apply the "no-changelog" label to this PR.

## References

NA

Issue #280, opened 2023-06-28
Auto-generate VRL function documentation


fuchsnj:
## Problem
Before the VRL repo was split from Vector, VRL function documentation was manually added for each change in CUE docs. Now since the repo is split, there is currently no way to add / update documentation with each PR. Even before, it was easy for documentation to become out of date since it's manually updated and nothing is checking it for accuracy.

## Proposal
The source of documentation will live wherever VRL functions are defined. (The VRL repo for stdlib functions, or other repos for user added functions, such as Vector specific VRL functions). This will be auto-generated directly from the code (function name, parameters, examples, .etc) and supplemented by annotations were needed. This will be roughly similar to Vector's [config schema](https://github.com/vectordotdev/vector/blob/master/rfcs/2022-02-23-9481-config-schema.md)

The VRL repo will contain a tool that can generate a computer-readable intermediate format of the function documentation in JSON. This allows end-users of VRL to generate documentation from the intermediate format.

For example:
- Vector would generate the intermediate JSON format for the stdlib functions, as well as several Vector specific functions, and then convert that to CUE docs for the documentation [here](https://vector.dev/docs/reference/vrl/functions/)
- VRL itself may one day host it's own documentation for just the stdlib. The intermediate format gives flexibility to use whatever tool makes sense for the job, instead of having to copy what Vector is currently using.
- The VRL playground (or Observability Pipelines) could ingest this data as they wish to improve the related UI.
- Future users of VRL could integrate this data into their own documentation.

pront:
Notes from our offline discussion:
* Utilizing Rust AST/macros and using the VRL code directly would be ideal  
* It would be great if all these `website/cue/reference/remap/functions` were generated purely based on VRL code.

pront:
Another caveat with the existing state of things is that you might get surprising failures when running VRL tests from within the Vector repo. That will run additional test cases, generated by docs examples.

PR #1613, opened 2026-01-26
feat!(stdlib): Add required optional default to Parameter

thomasqueirozb:
## Summary

Add required optional default to Parameter.

Part of the automatic documentation generation RFC (#280)

---

Last commit was generated with the following code:

```py
#!/usr/bin/env python3
"""
Convert the pattern:
    let x = self
        .field
        .as_ref()
        .map(|expr| expr.resolve(ctx))
        .transpose()?
        .unwrap_or_else(|| DEFAULT_*.clone());

to:
    let x = self
        .field
        .map_resolve_with_default(ctx, || DEFAULT_*.clone())?;
"""

import re
import sys
from pathlib import Path

# Pattern to match the old style with flexible whitespace
# self and .field_name are on separate lines
PATTERN = re.compile(
    r'self\s+\.(\w+)\s+'
    r'\.as_ref\(\)\s+'
    r'\.map\(\|expr\| expr\.resolve\(ctx\)\)\s+'
    r'\.transpose\(\)\?\s+'
    r'\.unwrap_or_else\(\|\| (DEFAULT_\w+\.clone\(\))\)',
    re.MULTILINE | re.DOTALL
)

def convert_file(filepath: Path) -> bool:
    """Convert a single file. Returns True if changes were made."""
    content = filepath.read_text()

    def replace(match):
        field_name = match.group(1)
        default = match.group(2)  # DEFAULT_*.clone()
        return f'self
            .{field_name}
            .map_resolve_with_default(ctx, || {default})?'

    new_content = PATTERN.sub(replace, content)

    if new_content != content:
        filepath.write_text(new_content)
        return True
    return False

def main():
    stdlib_dir = Path(__file__).parent.parent / 'src' / 'stdlib'

    if not stdlib_dir.exists():
        print(f"Error: {stdlib_dir} does not exist")
        sys.exit(1)

    changed_files = []
    for rs_file in sorted(stdlib_dir.glob('*.rs')):
        if convert_file(rs_file):
            changed_files.append(rs_file.name)
            print(f"Converted: {rs_file.name}")

    if changed_files:
        print(f"
Converted {len(changed_files)} files")
    else:
        print("No files needed conversion")

if __name__ == '__main__':
    main()
```
---

### Changes Checklist

**Defaults that differ from source code**
- decode_lz4.rs (default buf size differs from docs vs source code)
- encode_lz4.rs (default prepend_size differed from docs vs source code)
- format_number.rs (grouping_separator had wrong default "," in docs, actual behavior is no default)
- parse_bytes.rs (DEFAULT_BASE was Integer(2) but should be Bytes("2"))
- parse_etld.rs (DEFAULT_PLUS_PARTS was Boolean(false) but should be Integer(0))
- parse_groks.rs (DEFAULT_ALIASES was Boolean(true) but should be empty object)

**Changes that don't map 1:1 with source code**
- encode_zstd.rs (zstd level 0 means "use default 3", so docs and code are consistent)
- slice.rs (end has no runtime default, actual default is None meaning "to end")
- uuid_v7.rs (no runtime default, None means use Utc::now())

**No changes between docs and source code**
- ceil.rs
- compact.rs
- contains.rs
- crc.rs
- decode_base64.rs
- decode_punycode.rs
- del.rs
- encode_base64.rs
- encode_gzip.rs
- encode_key_value.rs
- encode_percent.rs
- encode_punycode.rs
- encode_zlib.rs
- ends_with.rs
- find.rs
- flatten.rs
- floor.rs
- format_int.rs
- from_unix_timestamp.rs
- hmac.rs
- log.rs
- map_keys.rs
- map_values.rs
- match_array.rs
- merge.rs
- parse_apache_log.rs
- parse_aws_alb_log.rs
- parse_common_log.rs
- parse_csv.rs
- parse_json.rs
- parse_key_value.rs
- parse_nginx_log.rs
- parse_regex_all.rs
- parse_regex.rs
- parse_url.rs
- parse_user_agent.rs
- parse_xml.rs
- remove.rs
- replace_with.rs
- replace.rs
- round.rs
- sha2.rs
- sha3.rs
- shannon_entropy.rs
- sieve.rs
- starts_with.rs
- to_unix_timestamp.rs
- unflatten.rs
- xxhash.rs

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Performance

## Is this a breaking change?

- [x] Yes
- [ ] No

## How did you test this PR?

`./scripts/checks.sh`

## References

- References: #1610

### DIFF
--- a/src/stdlib/parse_groks.rs
+++ b/src/stdlib/parse_groks.rs
@@ -72,6 +72,7 @@ use std::{fs::File, io::BufReader, path::Path};
 
 static DEFAULT_ALIASES: LazyLock<Value> =
     LazyLock::new(|| Value::Object(std::collections::BTreeMap::new()));
+static DEFAULT_ALIAS_SOURCES: LazyLock<Value> = LazyLock::new(|| Value::Array(vec![]));
 
 static PARAMETERS: LazyLock<Vec<Parameter>> = LazyLock::new(|| {
     vec![
@@ -101,7 +102,7 @@ static PARAMETERS: LazyLock<Vec<Parameter>> = LazyLock::new(|| {
             kind: kind::ARRAY,
             required: false,
             description: "Path to the file containing aliases in a JSON format.",
-            default: None,
+            default: Some(&DEFAULT_ALIAS_SOURCES),
         },
     ]
 });


src/stdlib/dns_lookup.rs
thomasqueirozb: This file is undocumented so the defaults weren't picked up automatically. This needs to be updated to use these defaults

src/stdlib/http_request.rs
thomasqueirozb: Same thing

src/stdlib/parse_cef.rs
```diff
@@ -62,12 +62,14 @@ impl Function for ParseCef {
...
                 required: true,
                 description: "The string to parse.",
+                default: None,
             },
             Parameter {
                 keyword: "translate_custom_fields",
                 kind: kind::BOOLEAN,
                 required: false,
                 description: "Toggles translation of custom field pairs to `key:value`.",
+                default: None,
```
thomasqueirozb: Should be false

src/stdlib/encode_json.rs
```diff
@@ -34,12 +34,14 @@ impl Function for EncodeJson {
...
                 required: true,
                 description: "The value to convert to a JSON string.",
+                default: None,
             },
             Parameter {
                 keyword: "pretty",
                 kind: kind::BOOLEAN,
                 required: false,
                 description: "Whether to pretty print the JSON string or not.",
+                default: None,
```
thomasqueirozb: Should be false

src/stdlib/parse_xml.rs
```diff
@@ -1,5 +1,77 @@
...
+            required: true,
+            description: "The string representation of the XML document to parse.",
+            default: None,
+        },
+        Parameter {
+            keyword: "trim",
+            kind: kind::BOOLEAN,
+            required: false,
+            description: "Remove excess whitespace between XML elements.",
+            default: None,
```
thomasqueirozb: Needs to be updated

src/stdlib/parse_nginx_log.rs
```diff
@@ -4,6 +4,40 @@ use regex::Regex;
 use std::collections::BTreeMap;
 
 use super::log_util;
+use std::sync::LazyLock;
+
+static DEFAULT_TIMESTAMP_FORMAT_STR: &str = "%d/%b/%Y:%T %z";
+static DEFAULT_TIMESTAMP_FORMAT: LazyLock<Value> =
+    LazyLock::new(|| Value::Bytes(Bytes::from(DEFAULT_TIMESTAMP_FORMAT_STR)));
```
thomasqueirozb, line 11: This is misleading, maybe it's best to have an explainer string as the default value instead

src/stdlib/encode_key_value.rs
```diff
@@ -29,6 +30,51 @@ pub(crate) fn encode_key_value(
...
+            required: true,
+            description: "The value to convert to a string.",
+            default: None,
+        },
+        Parameter {
+            keyword: "fields_ordering",
+            kind: kind::ARRAY,
+            required: false,
+            description: "The ordering of fields to preserve. Any fields not in this list are listed unordered, after all ordered fields.",
+            default: None,
```
thomasqueirozb: Technically an empty `vec![]`, could be consolidated into a default value here

src/stdlib/encode_logfmt.rs
```diff
@@ -21,12 +21,14 @@ impl Function for EncodeLogfmt {
...
                 required: true,
                 description: "The value to convert to a logfmt string.",
+                default: None,
             },
             Parameter {
                 keyword: "fields_ordering",
                 kind: kind::ARRAY,
                 required: false,
                 description: "The ordering of fields to preserve. Any fields not in this list are listed unordered, after all ordered fields.",
+                default: None,
```
thomasqueirozb: Same default should be used here (see encode_key_value)

src/stdlib/parse_etld.rs
```diff
@@ -2,8 +2,43 @@ use psl::Psl;
...
+            default: Some(&DEFAULT_PLUS_PARTS),
+        },
+        Parameter {
+            keyword: "psl",
+            kind: kind::BYTES,
+            required: false,
+            description: "Can be provided to use a different public suffix list.
+
+By default, https://publicsuffix.org/list/public_suffix_list.dat is used.",
+            default: Some(&DEFAULT_PSL),
```
thomasqueirozb: This is just plain wrong

src/stdlib/parse_groks.rs
```diff
@@ -66,9 +66,46 @@ mod non_wasm {
...
+            required: false,
+            description: "The shared set of grok aliases that can be referenced in the patterns to simplify them.",
+            default: Some(&DEFAULT_ALIASES),
+        },
+        Parameter {
+            keyword: "alias_sources",
+            kind: kind::ARRAY,
+            required: false,
+            description: "Path to the file containing aliases in a JSON format.",
+            default: None,
```
thomasqueirozb: Technically can be an empty array

bfung (approved):

PR #1649, opened 2026-02-19
feat(new function): implement an `encode_csv` VRL function

armleth:
## Summary

Adds a new `encode_csv` function to VRL. Currently, VRL includes `parse_csv` for decoding CSV data, but lacks the inverse operation to re-serialize arrays back to CSV. We already provide encoding functions for other formats like encode_json, encode_key_value, and encode_logfmt.

## Change Type

- [ ] Bug fix
- [X] New feature
- [ ] Performance

## Is this a breaking change?

- [ ] Yes
- [X] No

- [X] Yes. Please add a changelog fragment based on

## Checklist

- [X] Our [CONTRIBUTING.md](https://github.com/vectordotdev/vrl/blob/main/CONTRIBUTING.md) is a good starting place.
- [X] For new VRL functions, please also create a sibling PR in Vector to document the new function.


benches/stdlib.rs
```diff
@@ -1,11 +1,13 @@
-use chrono::{DateTime, Datelike, TimeZone, Utc};
-use criterion::{Criterion, criterion_group, criterion_main};
-use regex::Regex;
```
gwenaskell, line 3: It seems that your formatter is configured to group all imports in a single `use` statement. Can you configure it to avoid that? Since most of the files in that repo declare one `use` per root package, better to keep it consistent.

(note that grouping per-package, as you did for std and vrl below, is good IMO)

gwenaskell (commented):
Good job! Congrats 🎉 

Before merging this PR, I left one comment below, and we should also add a test VRL file in lib/tests/tests/functions. What you could do in that test is calling parse_csv to convert a csv row to a json object; encode it back to csv with this new function, and check that the output matches the initial value

src/stdlib/encode_csv.rs
```diff
@@ -0,0 +1,202 @@
...
+        TypeDef::bytes().fallible()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use {super::*, crate::value};
+
+    test_function![
+        parse_csv => EncodeCsv;
```
dd-sebastien-lb: encode_csv ? I don't think this impact the result of the test

benches/stdlib.rs
armleth: Done!

gwenaskell (approved):

src/stdlib/encode_csv.rs
```diff
@@ -0,0 +1,201 @@
...
+    test_function![
+        encode_csv => EncodeCsv;
+
+        valid {
+            args: func_args![value: value!(["foo", "bar", "foo \", bar"])],
+            want: Ok(value!("foo,bar,\"foo \"\", bar\"")),
+            tdef: TypeDef::bytes().fallible(),
+        }
+
+        invalid_utf8 {
```
pront, line 165: Do we want to allow invalid UTF8 bytes? Usually CSV works on plain text.

src/stdlib/encode_csv.rs
```diff
@@ -0,0 +1,201 @@
+use crate::{compiler::prelude::*, stdlib::csv_utils::parse_single_byte_delimiter};
+use csv::WriterBuilder;
+use std::sync::LazyLock;
+
+static DEFAULT_DELIMITER: LazyLock<Value> = LazyLock::new(|| Value::Bytes(Bytes::from(",")));
+
+static PARAMETERS: LazyLock<Vec<Parameter>> = LazyLock::new(|| {
+    vec![
+        Parameter::required("value", kind::ANY, "The value to convert to a CSV string."),
```
pront, line 9: Would be nice if this function worked on arrays of object as well. Is that something you considered supporting?

pront (approved):
Code LGTM. Left a couple of questions.

src/stdlib/encode_csv.rs
armleth: This test specifically comes from https://github.com/armleth/vrl/blob/0ca8b1061e3ebf9d66c797fafcbec0bc363fe5be/src/stdlib/parse_csv.rs#L152-L156`. The idea behind this is to be able to encode everything we can parse with `parse_csv` function. Wdyt about that?

gwenaskell: parse_csv seems to allow it, we should mirror its behavior IMO

src/stdlib/encode_csv.rs
gwenaskell: We needed a function that mirrors parse_csv, so we did not consider it. This sounds more complex because or the possibility of sparse keys across objects

OliviaShoup (approved):
nice catch! 🚀

PR #1680, opened 2026-03-03
feat(stdlib): gate crypto/hashing functions behind enable_crypto_functions feature

hligit:
## Summary

Add a new `enable_crypto_functions` cargo feature that controls whether cryptographic and hashing VRL stdlib functions are compiled into the binary.

When `enable_crypto_functions` is disabled, the following 12 functions are excluded at compile time: `md5`, `sha1`, `sha2`, `sha3`, `hmac`, `encrypt`, `decrypt`, `encrypt_ip`, `decrypt_ip`, `seahash`, `xxhash`, `crc`. Their 19 underlying crypto crate dependencies (`aes`, `sha-2`, `sha-3`, `hmac`, `digest`, `chacha20poly1305`, `crypto_secretbox`, etc.) are also excluded from the dependency tree.

The `redact` function's hash-based redaction modes (`sha2`, `sha3`) are also gated behind this feature. The function itself remains available with its `full` and `text` redactors. Attempting to use a hash redactor without the feature produces a clear error at config parse time.

This follows the existing pattern of `enable_network_functions`, `enable_system_functions`, and `enable_env_functions`. The feature is included in `default`, so existing users are unaffected.

### Motivation

Our organisation is pursuing a FIPS compliant build of Vector. FIPS 140-3 requires that all cryptographic operations use a CMVP-validated module (AWS-LC FIPS). VRL's stdlib includes hash and encryption functions implemented by independent Rust crates (`md-5`, `sha-1`, `crypto_secretbox`, etc.) that are not CMVP-validated. Being able to exclude these functions and their dependencies at compile time allows FIPS builds to guarantee that no non-validated crypto code is present in the binary.

Without this feature, the crypto crate dependencies are always compiled and linked, even if the functions are filtered at the application layer.

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Performance

## Is this a breaking change?

- [ ] Yes
- [x] No

`enable_crypto_functions` is included in `default` and was previously an implicit part of `stdlib`. Existing users who use `default` or `stdlib` features see no change. Users who construct custom feature sets and want crypto functions must now explicitly add `enable_crypto_functions`.

## How did you test this PR?

**With `enable_crypto_functions` enabled (default):**

```
cargo check
cargo test --lib -- redact
```

All 13 redact tests pass, including sha2/sha3 hash tests. All crypto stdlib functions are present.

**Without `enable_crypto_functions`:**

```
cargo check --no-default-features --features "compiler,value,diagnostic,path,parser,stdlib,datadog,core,enable_env_functions,enable_system_functions"
cargo test --no-default-features --features "compiler,value,diagnostic,path,parser,stdlib,datadog,core,enable_env_functions,enable_system_functions" --lib -- redact
```

Build succeeds. 11 redact tests pass (hash tests excluded). Two additional
tests verify that `sha2` and `sha3` redactors are rejected with an error
when the feature is disabled.

```
./scripts/checks.sh
```
All checks passed.

Change log:
```
Added a new `enable_crypto_functions` cargo feature (included in `default`) that gates 12 crypto/hashing stdlib functions and their dependencies. This allows consumers to exclude non-validated crypto code from builds that require a specific cryptographic backend (e.g., FIPS 140-3 compliance).
```

## Checklist


src/stdlib/ip_utils.rs
```diff
@@ -1,12 +1,10 @@
-use crate::compiler::ExpressionError;
-use crate::prelude::{VrlValueConvert, value::Value};
-use std::convert::TryInto;
-
+#[cfg(feature = "enable_crypto_functions")]
 pub(crate) fn to_key<const N: usize>(
```
hligit, line 5: This function is only called by `encrypt_ip` and `decrypt_ip`, putting them behind the feature gate to avoid compiler warnings.

src/stdlib/ip_utils.rs
```diff
@@ -1,12 +1,10 @@
-use crate::compiler::ExpressionError;
```
pront, line 1: Nit: could use cfg_if! to avoid

src/stdlib/redact.rs
```diff
@@ -619,6 +635,67 @@ mod test {
             tdef: TypeDef::bytes().infallible(),
         }
 
+        invalid_redactor {
+             args: func_args![
+                 value: "hello 123456 world",
+                 filters: vec!["us_social_security_number"],
+                 redactor: "not a redactor"
+             ],
+             want: Err("invalid argument"),
```
pront, line 644: Could use a better error message here e.g. `"sha2/sha3 redactors require the enable_crypto_functions feature"`

Cargo.toml
pront, line 84: enable_crypto_functions should be included here but now I realize we want to include all `enable_*_functions` features here. I will submit a PR to address this shortly and you can rebase later.

pront (commented):
Thanks, this is a reasonable feature / use case.

src/stdlib/redact.rs
hligit: From my understanding, the `test_function!` macro only compares [error mesasges](https://github.com/vectordotdev/vrl/blob/43a08e226c32bdc1b51f9471c47839c068854b85/src/compiler/test_util.rs#L104).  In order to propagate a different error message, a new [Function::Error](https://github.com/vectordotdev/vrl/blob/43a08e226c32bdc1b51f9471c47839c068854b85/src/compiler/function.rs#L625) variant need to be introduced. I'm not 100% sure this is what you wanted?

Another way is instead of the test_function! macro, use Redact::compile directly and compare the error labels. That also seem to be overkill for this.

Or do you have a better idea how this should be implemented?

src/stdlib/ip_utils.rs
hligit: Done

src/stdlib/redact.rs
hligit: Please ignore above. The test_function! macro already handles this.
I've updated the error message as you suggested.

Cargo.toml
hligit: Done!

src/stdlib/ip_utils.rs
```diff
@@ -1,16 +1,19 @@
...
-use crate::prelude::{VrlValueConvert, value::Value};
-use std::convert::TryInto;
+cfg_if::cfg_if! {
```
pront: I realized we can just do

```rust
#[cfg(feature = "enable_crypto_functions")]
mod ip_utils;
```

in `src/stdlib/mod.rs`

pront (approved):
One thing: I don't see the updated error message

src/stdlib/ip_utils.rs
hligit: Good idea! Fixed.

pront (approved):

PR #1741, opened 2026-04-17
chore(releasing): rewrite release tooling in Rust

pront:
## Summary

This replaces the release tooling with a native Rust crate, eliminating the Python ecosystem dependency (scripts, pip packages, towncrier) from the release process.

The release flow is also simplified from a two-step process (create PR → wait for merge → publish) to a single interactive command: `cargo run -p release`. It bumps the version, generates the changelog, pauses for review, publishes to crates.io, tags, and opens a PR to land the changes into main.

### Tradeoffs

**What we gain:**
- Zero non-Rust dependencies for the release process
- Single command instead of a multi-step workflow
- Changelog assembly is ~80 lines of Rust we own, vs a transitive Python dependency tree (towncrier)
- CI changelog validation compiles with the workspace — no separate Python install step

**What we give up:**
- We now own the changelog assembly logic. towncrier handled edge cases we might not have hit (e.g., unusual markdown in fragments). Low risk given our simple fragment format.
- First `cargo run -p release` on a clean checkout needs to compile (~4s). Cached builds are instant.
- The single-step flow means `cargo publish` runs before the PR is merged. If publish succeeds but the PR merge somehow fails, the crate is on crates.io but main doesn't reflect the release yet. This is recoverable (just merge the PR), and the old flow had the inverse risk (PR merged but someone forgets to run `publish.py`).

**New workspace dependencies (release crate only):** `semver`, `toml_edit`, `ureq`, `chrono`.

### Usage

```shell
cargo run -p release              # bump minor (default), full flow
cargo run -p release -- major     # bump major
cargo run -p release -- patch     # bump patch
cargo run -p release -- 1.2.3     # exact version
cargo run -p release -- --dry-run # preview changelog, no changes
cargo run -p release -- check-changelog  # CI: validate fragment filenames
```

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## Is this a breaking change?

- [ ] Yes
- [x] No

## How did you test this PR?

- `cargo check -p release` — compiles cleanly
- `cargo clippy -p release -- -D warnings` — no warnings
- `cargo fmt --check -p release` — formatted

- [ ] Yes.

## References

Replaces `release/*.py`, `scripts/generate_release_changelog.sh`, `scripts/check_changelog_fragments.sh`, and `changelog.toml` with the `release/` Rust crate.


release/README.md
```diff
@@ -1,15 +1,43 @@
...
+
+### Release (default)
 
 ```shell
-python3 create_release_pull_request.py <version>
+# Bump minor version (default), generate changelog, publish, tag, create PR
+cargo run -p release
+
+# Bump a specific component
+cargo run -p release -- major
```
thomasqueirozb, line 14: I don't think we should support bumping a major version this easily. I think we need to have an additional flag here so we never to do by mistake - or just completely remove this functionality

release/README.md
```diff
@@ -1,15 +1,43 @@
...
+cargo run -p release -- patch
+
+# Exact version
+cargo run -p release -- 1.2.3
```
thomasqueirozb, line 18: Related to the above, we should add validation that we're not bumping a major version or skipping minor versions by mistake

release/README.md
pront: I'd push back on this:

1. It's already explicit — you have to type `cargo run -p release -- major`. The default (no argument) is a minor bump, so there's no path where you accidentally trigger a major release.
2. The old Python script had the same behavior — `python3 create_release_pull_request.py major` worked identically with no confirmation flag. This isn't a new capability.
3. The interactive pause before publish gives you a clear chance to ctrl-c if something looks wrong. Adding a `--confirm` flag on top of that is redundant complexity for a scenario that hasn't happened.

If we want to guard against fat-finger mistakes, the existing flow already does: you see "New version: 1.0.0", you see the changelog preview, and you have to press Enter to continue. That's three opportunities to bail.

release/README.md
pront: Good call. Added validation in `7718194fd` — when an exact version is passed, it now checks that it's a valid single-step bump from current (major+1, minor+1, or patch+1). Skipped versions, downgrades, and same-version are all rejected with a clear error showing the three valid options.

```
$ cargo run -p release -- 5.0.0
Error: Version 5.0.0 is not a valid bump from 0.32.0. Expected one of: 1.0.0 (major), 0.33.0 (minor), 0.32.1 (patch).
```

release/README.md
thomasqueirozb: True, I agree it's a non issue considering we create a PR from this and it is not fully automated. Resolving

thomasqueirozb (approved):
Nice! 🐍 🪓

PR #1760, opened 2026-05-04
enhancement(compiler): surface multiple unhandled errors in one compilation

pront:
## Motivation

The compiler tracked at most one pending unhandled fallibility per root expression via `fallible_expression_error: Option<CompilerError>`. This had two costs the previous fix (#1758) flagged but didn't address:

1. **Lost diagnostics under hard errors.** When `Assignment::new` itself returned an error mid-compile, any prior pending fallibility was dropped — the user would fix the surfaced error, recompile, and *then* discover the next one.
2. **Manual save/restore at every callsite.** `compile_op` and `compile_assignment` had to take/restore the slot just so a sub-expression couldn't accidentally clobber a peer's pending error.

## Summary

Field becomes `fallible_expressions: Vec<CompilerError>` — a stack of pending unhandled fallibilities. Each helper snapshots `len()` on entry and leaves prior entries untouched. Consumers (`??`, infallible-typed op result, `x, err =`, assignment success or hard failure) truncate back to their own snapshot. The root-expression boundary drains every still-pending entry into the diagnostic list, even when the root itself failed to compile.

### User-visible change

Programs that compile today still compile; only the error path changes. Programs that fail can now surface multiple independent errors in one pass instead of one-error-at-a-time:

```vrl
{
    push(.x, 1)
    .b = push(.y, 2)
}
```

Now reports:

```
error[E110]: invalid argument type
  ┌─ :2:10
  │
2 │     push(.x, 1)

error[E103]: unhandled fallible assignment
  ┌─ :3:10
  │
3 │     .b = push(.y, 2)
```

Previously only the E103 surfaced; the user fixed it, recompiled, and only then saw the E110.

The implementation is careful about cascade noise: when an assignment hard-errors on a fallible RHS, the RHS-pushed entry is consumed by the assignment error rather than re-flushed, so the user doesn't see *both* "RHS is fallible" *and* "assignment of fallible RHS" — just the actionable one.

## Change Type

- [x] New feature

## Is this a breaking change?

- [x] No

## How did you test this PR?

- New regression test `multiple_unhandled_fallibilities_surface_in_one_pass` covering the previously-dropped path.
- Existing #453/#36 regression tests still pass.
- Full library suite (1836 tests), `cargo fmt`, `./scripts/clippy.sh` clean.
- Manually verified the worked example and additional shapes (closure body with two discarded fallibles; both branches of an `if`/`else`).

- [x] Yes (changelog fragment included).

## References

- Follow-up to #1758 — implements the TODO it left in `compile_assignment`.

### DIFF
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -46,15 +46,20 @@ pub struct Compiler<'a> {
     /// errors when the reason for it being undefined is another compiler error.
     skip_missing_query_target: Vec<(QueryTarget, OwnedValuePath)>,
 
-    /// Track which expression in a chain of expressions is fallible.
+    /// Stack of unhandled fallibility errors collected while compiling the
+    /// current root expression.
     ///
-    /// It is possible for this state to switch from `None`, to `Some(T)` and
-    /// back to `None`, if the parent expression of a fallible expression
-    /// nullifies the fallibility of that expression.
-    // This should probably be kept on the call stack as the "compile_*" functions are called
-    // otherwise some expressions may remove it when they shouldn't (such as the RHS of an operation removing
-    // the error from the LHS)
-    fallible_expression_error: Option<CompilerError>,
+    /// Each entry is the error produced by a fallible expression whose
+    /// fallibility has not (yet) been consumed by a parent (assignment,
+    /// error-coalesce `??`, etc.). Helpers track their own boundary by
+    /// snapshotting `len()` on entry and either leaving sub-expression
+    /// entries pending, or truncating back to the snapshot to consume them.
+    /// At the root expression boundary any entries still pending are
+    /// flushed as diagnostics. Using a stack instead of a single slot lets
+    /// multiple unhandled errors surface in one compilation pass — e.g.
+    /// when both a discarded fallible call and a later fallible assignment
+    /// exist in the same block.
+    pending_fallibilities: Vec<CompilerError>,
 
     config: CompileConfig,
 }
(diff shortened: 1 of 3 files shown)


src/compiler/expression/assignment.rs
```diff
@@ -981,31 +961,73 @@ mod test {
...
+    /// (E651). The rhs's fallibility is subsumed by the op-level error
+    /// and must not double-report as E100 at the root drain.
+    #[test]
+    fn op_hard_error_drains_subexpression_fallibilities() {
+        assert_eq!(error_codes("1 ?? parse_json(\"{}\")"), vec![651]);
+    }
+
+    /// Compile `src` and return the error codes in the order they were
+    /// emitted as diagnostics. Panics if compilation succeeds.
+    fn error_codes(src: &str) -> Vec<usize> {
```
thomasqueirozb: Nit: I rather use `error_codes_and_spans` everywhere instead of `error_codes` so that any span-level regressions are also caught

src/compiler/compiler.rs
```diff
@@ -419,35 +453,41 @@ impl<'a> Compiler<'a> {
...
+
         let lhs_span = lhs.span();
         let lhs = Node::new(lhs_span, self.compile_expr(*lhs, state)?);
 
-        // If we're using error-coalescing, we need to negate any tracked
-        // fallibility error state for the lhs expression.
+        // `??` consumes any fallibility produced by the lhs.
         if opcode.inner() == &Opcode::Err {
-            self.fallible_expression_error = None;
+            self.pending_fallibilities.truncate(pre_op_pending);
```
thomasqueirozb, line 466: nit: I'm not the biggest fan of index-based truncating as it hurts readability and can introduce bugs. Having a pending fallibility context or using pending fallibility as a local array instead of a struct level property may be an improvement that could improve readability. This is probably better suited as a follow up and not in this PR

thomasqueirozb (dismissed):

thomasqueirozb (approved):

src/compiler/expression/assignment.rs
pront: https://github.com/vectordotdev/vrl/pull/1760/commits/b7063c6dc54d7552806d11b85e11cab650f1881c

PR #1762, opened 2026-05-06
fix(protobuf): support non-string map keys in encode_proto and parse_…

flaviofcruz:
## Summary

Protobuf permits map keys of any scalar type (bool, int32, int64, uint32, uint64, string), but VRL's `ObjectMap` only allows string keys. Both `encode_proto` and `parse_proto` previously assumed `MapKey::String`, which caused:

- **`parse_proto`**: failed with `Internal error while parsing protobuf map. Field descriptor: ...` for any map with non-string keys (e.g. `map<int64, bool>`).
- **`encode_proto`**: silently dropped the entire field with a warning when `try_set_field` rejected the wrong `MapKey` variant.

 ## Fix

Convert non-string map keys to/from their decimal string representation on both sides:

- `parse_proto` stringifies `Bool` / `I32` / `I64` / `U32` / `U64` keys (`v.to_string()`).
- `encode_proto` looks up the key field's `Kind` from the map entry descriptor (via `map_entry_key_field()`) and parses the string back into the appropriate `MapKey` variant via a small `parse_map_key(&Kind, &str) -> Result<MapKey, String>` helper.

Both sides cover all four wire encodings per integer width (`Int32`/`Sint32`/`Sfixed32` → `i32`, etc.).

## Test plan

- [x] New `test_protobuf_maps` fixture (`.proto` + `.desc`) defining a `Maps` message with one `map<K, string>` field per supported proto key type (`by_bool`, `by_int32`, `by_int64`, `by_uint32`, `by_uint64`, `by_string`).
- [x] `test_encode_decode_map_all_key_types` round-trips a VRL `Object` containing all six map shapes through `encode_proto` → `parse_proto` and asserts equality. Single test exercises both new code paths.
- [x] `cargo fmt --check --all` clean.
- [x] `cargo clippy --all-targets --features enable_system_functions -- -D warnings` clean.
- [x] All existing `protobuf::*` tests still pass (20/20).


tests/data/protobuf/test_protobuf_maps/v1/test_protobuf_maps.proto
```diff
@@ -0,0 +1,15 @@
+syntax = "proto3";
+
+package test_protobuf_maps.v1;
+
+// Fixture exercising every supported proto map key type. Used by
+// encode_proto/parse_proto round-trip tests to verify that non-string
+// keys are correctly converted to/from their decimal string form.
+message Maps {
+  map<bool,   string> by_bool   = 1;
```
pront, line 9: This has some interesting implications around key parsing but a map from bool to anything should be something we rarely (never?) see in practice.

pront (approved):
Looks good. I updated the changelog. Thank you for this nice fix!

PR #1545, opened 2025-10-29
enhancement(protobuf encoding): support int to double conversion

thomasqueirozb:
## Summary

As title states

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Non-functional (chore, refactoring, docs)
- [ ] Performance

## Is this a breaking change?

- [ ] Yes
- [x] No

## How did you test this PR?

Ran E2E otel Vector test with telemetrygen generated metric with type histogram

## Does this PR include user facing changes?

- [ ] Yes. Please add a changelog fragment based on
  our [guidelines](https://github.com/vectordotdev/vector/blob/master/changelog.d/README.md).
- [x] No. A maintainer will apply the "no-changelog" label to this PR.

## Checklist

- [ ] Our [CONTRIBUTING.md](https://github.com/vectordotdev/vrl/blob/main/CONTRIBUTING.md) is a good starting place.
- [ ] If this PR introduces changes to [LICENSE-3rdparty.csv](https://github.com/vectordotdev/vrl/blob/main/LICENSE-3rdparty.csv), please
  run `dd-rust-license-tool write` and commit the changes. More details [here](https://crates.io/crates/dd-rust-license-tool).
- [ ] For new VRL functions, please also create a sibling PR in Vector to document the new function.

PR #1763, opened 2026-05-06
enhancement(protobuf encoding): coerce integer/string into bool and integer into float

flaviofcruz:
Adds three coercions to `encode_proto`'s `convert_value_raw`, mirroring
  the existing `Integer -> Double` conversion (#1545):

  - `(Value::Integer, Kind::Bool) => Bool(i != 0)`
  - `(Value::Bytes, Kind::Bool)` — use the function underlying VRL's `to_bool` to convert string to bool.
  - `(Value::Integer, Kind::Float) => F32(i as f32)` — the obvious counterpart to the existing `Integer -> Double` arm.

  ## Why

  `encode_proto` already accepts integers-from-bytes and bytes-as-numerics
  for every numeric proto scalar (Int32/Int64/Sint32/Float/Double/...).
  Bool fields were the odd one out: writing `42` or `"true"` into a
  `bool` would fail the encode even though equivalent inputs succeed for
  every neighboring scalar type. This patch closes that gap.

  The `Integer -> Float` arm is symmetric with `Integer -> Double` and is
  trivially safe (the `as f32` cast already happens elsewhere in this file
  for `Float -> Float`).

  ## Test plan

  - [x] New `Booleans` message in `tests/data/protobuf/test/v1/test.proto`.
  - [x] `test_encode_integer_as_float` (mirrors the existing `..._as_double`).
  - [x] `test_encode_integer_as_bool` covers `0/1/42/-1`.
  - [x] `test_encode_string_as_bool` covers all true/false vocabularies plus
        asserts that empty/invalid strings return an encode error.
  - [x] `cargo fmt --check --all` clean.
  - [x] `cargo clippy --all-targets --features enable_system_functions -- -D warnings` clean.
  - [x] All `protobuf::*` tests pass (22/22).


src/protobuf/encode.rs
```diff
@@ -22,6 +22,15 @@ fn convert_value_raw(
     let kind_str = value.kind_str().to_owned();
     match (value, kind) {
         (Value::Boolean(b), Kind::Bool) => Ok(prost_reflect::Value::Bool(b)),
+        (Value::Integer(i), Kind::Bool) => Ok(prost_reflect::Value::Bool(i != 0)),
+        (Value::Bytes(b), Kind::Bool) => {
```
pront: Here I would use `Conversion::Boolean.convert(b)` because we already have string→bool conversion semantic in VRL (`to_bool`).

pront (commented):
Thanks @flaviofcruz, please add a changelog since this is a user facing change.

src/protobuf/encode.rs
flaviofcruz: Done!

pront (approved):

PR #1772, opened 2026-05-20
perf(parse_regex, parse_regex_all): pre-compute capture names

thomasqueirozb:
## Summary

`parse_regex` does two things per event that can be eliminated. First, it reconstructs the map keys by calling `regex.capture_names()` on every event and allocating a fresh `KeyString` for each. Since the regex is compiled once at VRL compile time when the pattern is static, the names are fixed and can be stored then instead. Second, each captured substring is heap-copied into an owned `Bytes`. The input value is already a refcounted `Bytes` buffer, so captures can be zero-copy slices of it. We now get an `Arc` increment rather than a `memcpy`.

### Benchmarks (`cargo bench -- parse_regex`)

| Benchmark | `origin/main` | this branch | delta |
|---|---|---|---|
| `parse_regex/matches` | 3.061 µs | 2.660 µs | **−13%** |
| `parse_regex/single_match` | 260.9 ns | 233.2 ns | **−11%** |
| `parse_regex_all/matches` | 27.76 µs | 28.51 µs | +3% (noise) |

`parse_regex_all` shows no regression — the benchmark only has two short matches, so per-event overhead savings are small relative to total time.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Performance

## Is this a breaking change?

- [ ] Yes
- [x] No

## How did you test this PR?

Existing unit tests for `parse_regex` and `parse_regex_all` pass unchanged.

`cargo bench --bench stdlib --features="default test" -- --baseline main parse_regex`

## Checklist

## References

NA

### DIFF
--- a/src/stdlib/parse_regex.rs
+++ b/src/stdlib/parse_regex.rs
@@ -25,6 +25,20 @@ contains the whole match.",
     ]
 });
 
+fn parse_regex(
+    bytes: &bytes::Bytes,
+    pattern: &Regex,
+    capture_names: &[KeyString],
+    numeric_groups: bool,
+) -> Resolved {
+    util::with_utf8_bytes(bytes, |s, utf8_bytes| {
+        let parsed = pattern.captures(s).map(|capture| {
+            util::capture_regex_to_map(&capture, capture_names, numeric_groups, utf8_bytes)
+        });
+        Ok(parsed.ok_or("could not find any pattern matches")?.into())
+    })
+}
+
 #[derive(Clone, Copy, Debug)]
 pub struct ParseRegex;
 
@@ -152,20 +166,6 @@ impl Function for ParseRegex {
             },
         ]
     }
-}
-
-fn parse_regex(
-    bytes: &bytes::Bytes,
-    pattern: &Regex,
-    capture_names: &[KeyString],
-    numeric_groups: bool,
-) -> Resolved {
-    util::with_utf8_bytes(bytes, |s, utf8_bytes| {
-        let parsed = pattern.captures(s).map(|capture| {
-            util::capture_regex_to_map(&capture, capture_names, numeric_groups, utf8_bytes)
-        });
-        Ok(parsed.ok_or("could not find any pattern matches")?.into())
-    })
 }
 
 #[derive(Debug, Clone)]
--- a/src/stdlib/parse_regex.rs
+++ b/src/stdlib/parse_regex.rs
@@ -161,9 +161,9 @@ fn parse_regex(
     numeric_groups: bool,
 ) -> Resolved {
     util::with_utf8_bytes(bytes, |s, utf8_bytes| {
-        let parsed = pattern
-            .captures(s)
-            .map(|capture| util::capture_regex_to_map(&capture, capture_names, numeric_groups, utf8_bytes));
+        let parsed = pattern.captures(s).map(|capture| {
+            util::capture_regex_to_map(&capture, capture_names, numeric_groups, utf8_bytes)
+        });
         Ok(parsed.ok_or("could not find any pattern matches")?.into())
     })
 }
--- a/src/stdlib/parse_regex.rs
+++ b/src/stdlib/parse_regex.rs
@@ -160,17 +160,12 @@ fn parse_regex(
     capture_names: &[KeyString],
     numeric_groups: bool,
 ) -> Resolved {
-    let parsed = if let Ok(s) = std::str::from_utf8(bytes) {
-        pattern.captures(s).map(|capture| {
-            util::capture_regex_to_map(&capture, capture_names, numeric_groups, Some(bytes))
-        })
-    } else {
-        let s = String::from_utf8_lossy(bytes);
-        pattern.captures(s.as_ref()).map(|capture| {
-            util::capture_regex_to_map(&capture, capture_names, numeric_groups, None)
-        })
-    };
-    Ok(parsed.ok_or("could not find any pattern matches")?.into())
+    util::with_utf8_bytes(bytes, |s, utf8_bytes| {
+        let parsed = pattern
+            .captures(s)
+            .map(|capture| util::capture_regex_to_map(&capture, capture_names, numeric_groups, utf8_bytes));
+        Ok(parsed.ok_or("could not find any pattern matches")?.into())
+    })
 }
 
 #[derive(Debug, Clone)]
--- a/src/stdlib/parse_regex.rs
+++ b/src/stdlib/parse_regex.rs
@@ -1,5 +1,5 @@
 use crate::compiler::prelude::*;
-use crate::value::KeyString;
+use crate::value::{KeyString, ObjectMap};
 use regex::Regex;
 
 use super::util;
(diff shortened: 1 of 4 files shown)


src/stdlib/util.rs
```diff
@@ -27,32 +27,63 @@ where
...
+        (s, bytes)
+    } else {
+        owned = bytes::Bytes::from(String::from_utf8_lossy(bytes).into_owned());
+        (
+            std::str::from_utf8(&owned).expect("lossy string is valid UTF-8"),
+            &owned,
+        )
+    };
+    f(s, utf8_bytes)
+}
```
pront: This is a awkward implementation, can do a `match std::str::from_utf8(bytes) {` and take it from there?

pront (dismissed):
The PR description doesn't describe the optimization idea.

src/stdlib/util.rs
```diff
@@ -27,32 +27,63 @@ where
...
+    utf8_bytes: &bytes::Bytes,
 ) -> ObjectMap {
-    let names = regex.capture_names().flatten().map(|name| {
-        (
-            name.to_owned().into(),
-            capture.name(name).map(|s| s.as_str()).into(),
-        )
+    let names = capture_names.iter().map(|name| {
+        let value: Value = match capture.name(name.as_str()) {
+            Some(m) => utf8_bytes.slice(m.start()..m.end()).into(),
```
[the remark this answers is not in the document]
thomasqueirozb: ~Events are short lived and memory pressure shouldn't be an issue here. Deallocation would happen shortly after this runs.~

src/stdlib/util.rs
thomasqueirozb: It is unfortunately awkward but using match give me a clippy warning: https://rust-lang.github.io/rust-clippy/rust-1.92.0/index.html#single_match_else

This pattern looks off because there is no way to get the inner bytes from `String::from_utf8_lossy` and also keep a reference to the underlying `&str`. Will keep as is but open to suggestions as I'm not the biggest fan of this either

pront (commented):
Claude findings (second pass after the latest changes).

src/stdlib/util.rs
```diff
@@ -27,32 +27,63 @@ where
...
-/// Any named captures are also added to the Map with the key as the name.
+/// When `bytes` is valid UTF-8 both references point to the original buffer.
+/// Otherwise a lossy copy is allocated and both references point to that copy,
+/// keeping the regex byte offsets aligned with the buffer used for slicing.
+/// The closure receives `(s, utf8_bytes)` and its return value is forwarded.
+pub(crate) fn with_utf8_bytes<F, T>(bytes: &bytes::Bytes, f: F) -> T
+where
+    F: FnOnce(&str, &bytes::Bytes) -> T,
+{
+    let owned;
```
pront: **Claude finding (1):** the `let owned;` late-init is only needed to make both arms of the `if/else` produce the same `(&str, &Bytes)` tuple. A `match` with each arm calling `f(..)` directly would drop the late binding and read more straightforwardly:

```rust
match std::str::from_utf8(bytes) {
    Ok(s) => f(s, bytes),
    Err(_) => {
        let owned = bytes::Bytes::from(String::from_utf8_lossy(bytes).into_owned());
        let s = std::str::from_utf8(&owned).expect("from_utf8_lossy yields valid UTF-8");
        f(s, &owned)
    }
}
```

Both shapes are correct — current one is just slightly unusual.

src/stdlib/util.rs
```diff
@@ -27,32 +27,63 @@ where
...
+    let (s, utf8_bytes): (&str, &bytes::Bytes) = if let Ok(s) = std::str::from_utf8(bytes) {
+        (s, bytes)
+    } else {
+        owned = bytes::Bytes::from(String::from_utf8_lossy(bytes).into_owned());
+        (
+            std::str::from_utf8(&owned).expect("lossy string is valid UTF-8"),
```
pront: **Claude finding (5):** no test exercises the lossy fallback path — the entire `else` branch (including this `expect`) is uncovered. Worth adding at least one case feeding invalid UTF-8 (e.g. `Bytes::from_static(b"\xff foo bar")`) through `parse_regex`/`parse_regex_all` so the offset-alignment invariant this helper exists for is regression-tested.

src/stdlib/util.rs
```diff
@@ -27,32 +27,63 @@ where
...
+
+/// Fills an [`ObjectMap`] from a regex [`Captures`](regex::Captures).
+///
+/// Named captures are inserted under their group name; numeric groups (when
+/// `numeric_groups` is `true`) are inserted under their zero-based index, with
+/// `"0"` holding the full match.
 ///
+/// `capture_names` must be the pre-computed slice of named-group
+/// [`KeyString`]s for the regex (computed once at VRL compile time via
+/// `regex.capture_names().flatten().map(KeyString::from)`).
```
pront: **Claude finding (4):** the "computed once at VRL compile time" claim is true for `parse_regex` but not for `parse_regex_all`, which recomputes per `resolve()` when the pattern is not a compile-time constant. Suggest softening to something like: "computed at compile time when the regex is a constant, otherwise once per `resolve` call."

src/stdlib/parse_regex_all.rs
```diff
@@ -165,15 +184,27 @@ impl FunctionExpression for ParseRegexAllFn {
...
             .resolve(ctx)?
             .as_regex()
             .ok_or_else(|| ExpressionError::from("failed to resolve regex"))?
             .clone();
+        let bytes = value.try_bytes()?;
+
+        let capture_names: &[KeyString] = if let Some(names) = &self.capture_names {
+            names.as_slice()
+        } else {
+            &pattern
```
pront: **Claude finding (2):** this relies on temporary lifetime extension in a way that looks like a dangling reference. `&pattern.capture_names()...collect::<Vec<_>>()` borrows a `Vec` that has no `let` binding; it stays alive only because temporary lifetime extension on `let` keeps it around for the rest of `resolve()`. Compiles and works, but most readers will pause.

Either hoist explicitly:

```rust
let computed;
let capture_names: &[KeyString] = match &self.capture_names {
    Some(names) => names.as_slice(),
    None => {
        computed = pattern.capture_names().flatten().map(KeyString::from).collect::<Vec<_>>();
        &computed
    }
};
```

…or push the `parse_regex_all(..)` call into each branch so no shared binding is needed.

src/stdlib/parse_regex.rs
```diff
@@ -97,11 +103,17 @@ impl Function for ParseRegex {
     ) -> Compiled {
         let value = arguments.required("value");
         let pattern = arguments.required_regex("pattern", state)?;
+        let capture_names: Vec<KeyString> = pattern
+            .capture_names()
+            .flatten()
+            .map(KeyString::from)
+            .collect();
```
pront, line 137: **Claude finding (3):** `regex.capture_names().flatten().map(KeyString::from).collect()` is duplicated across three call sites — here, `parse_regex_all.rs` compile() (lines 102-109), and `parse_regex_all.rs` resolve() (lines 200-204). Extract a small helper in `util.rs`:

```rust
pub(crate) fn capture_names(regex: &Regex) -> Vec<KeyString> {
    regex.capture_names().flatten().map(KeyString::from).collect()
}
```

PR #1788, opened 2026-06-01
perf(parse_regex, parse_regex_all): literal regex in parse_regex/parse_regex_all and fix performance regression

thomasqueirozb:
## Summary

When the regex pattern is a compile-time literal, `parse_regex` and `parse_regex_all` now clone the `Regex` and hold it by value in a `PatternSource::Static` variant rather than keeping it as a `Box<dyn Expression>`. `regex::Regex::clone` allocates a fresh `Pool` per clone, so each Tokio worker that holds a copy of the function gets its own owner-thread fast path. Previously all workers shared a single `Arc<Regex>` (and therefore a single `Pool`), forcing all but one through `Pool::get_slow` under concurrent load.

Dynamic patterns (non-literal regex arguments) continue to resolve through the `PatternSource::Dynamic` path with no behavioral change.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Performance

## Is this a breaking change?

- [ ] Yes
- [x] No

## How did you test this PR?

Existing test suite covers correctness. The performance improvement is mechanical: each clone of the compiled function now gets its own `Pool`.

## Checklist

## References

NA

### DIFF
--- a/src/stdlib/parse_regex.rs
+++ b/src/stdlib/parse_regex.rs
@@ -103,16 +103,31 @@ impl Function for ParseRegex {
         arguments: ArgumentList,
     ) -> Compiled {
         let value = arguments.required("value");
-        let pattern = arguments.required("pattern");
-        let capture_info = pattern.resolve_constant(state).and_then(|v| {
-            v.as_regex().map(|r| {
-                r.capture_names()
+        let pattern_expr = arguments.required("pattern");
+        let numeric_groups = arguments.optional("numeric_groups");
+
+        // When the pattern is a literal regex, clone the `Regex` out and store it
+        // by value. `regex::Regex::clone()` allocates a fresh `Pool` per clone
+        // (only the compiled state is Arc-shared), so each clone of `ParseRegexFn`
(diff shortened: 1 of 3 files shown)


src/stdlib/parse_regex.rs
```diff
@@ -172,10 +187,20 @@ impl Function for ParseRegex {
     }
 }
 
+#[derive(Debug, Clone)]
+enum PatternSource {
```
JakubOnderka, line 191: I am using similar approach in `ip_cidr_contains` function, but I used names Constant and Expression. I think we should be consistent. What do you think is better? I am OK with changing names in  `ip_cidr_contains`.

https://github.com/vectordotdev/vrl/blob/f67b91c37fa94f5375a82081120c5b96b2e01247/src/stdlib/ip_cidr_contains.rs#L168-L171

thomasqueirozb: Constant/Expression is indeed a better choice. Good catch!

thomasqueirozb:
Due to #1789 I'm closing this PR and extracted the relevant changes into #1798 for `parse_regex_all` only for the time being.

PR #1797, opened 2026-06-03
test(stdlib): gate dns_lookup and reverse_dns network tests behind the test feature

thomasqueirozb:
## Summary

`dns_lookup` and `reverse_dns` tests were making live external DNS queries (to `google.com`, `dns.google`, `8.8.8.8`, etc.) on every test run. These are now gated behind the existing `test` feature flag, which is already documented as the mechanism for tests with external network dependencies.

`invalid_option_type` and `negative_int_type` in `dns_lookup` are unchanged since they fail at option validation before any DNS call is made.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## Is this a breaking change?

- [ ] Yes
- [x] No

## How did you test this PR?

`cargo test -p vrl --lib --features stdlib` passes clean (1845 tests, 0 failures).

## Checklist

## References

NA

### DIFF
--- a/src/stdlib/dns_lookup.rs
+++ b/src/stdlib/dns_lookup.rs
@@ -692,9 +692,12 @@ impl Function for DnsLookup {
 #[cfg(test)]
 #[cfg(not(target_arch = "wasm32"))]
 mod tests {
-    use std::collections::{BTreeMap, HashSet};
+    use std::collections::BTreeMap;
+    #[cfg(feature = "test")]
+    use std::collections::HashSet;
 
     use super::*;
+    #[cfg(feature = "test")]
     use crate::value;
 
     impl Default for DnsLookupFn {
@@ -708,6 +711,7 @@ mod tests {
         }
     }
 
+    #[cfg(feature = "test")]
     #[test]
     fn test_invalid_name() {
         let result = execute_dns_lookup(&DnsLookupFn {
@@ -728,6 +732,7 @@ mod tests {
         );
     }
 
+    #[cfg(feature = "test")]
     #[test]
     #[cfg(target_os = "linux")]
     // MacOS resolver doesn't always handle localhost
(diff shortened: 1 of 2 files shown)


src/stdlib/dns_lookup.rs
```diff
@@ -692,9 +692,12 @@ impl Function for DnsLookup {
 #[cfg(test)]
```
pront, line 692: We should add the `#[cfg(feature = "test")]` gate here once (vs above each test).

src/stdlib/reverse_dns.rs
pront, line 105: I would be fine with doing [this](https://github.com/vectordotdev/vrl/pull/1797/changes#r3349899043) here as well for all tests.

src/stdlib/dns_lookup.rs
thomasqueirozb: invalid_option_type and negative_int_type do not depend on the network, all others do

src/stdlib/reverse_dns.rs
thomasqueirozb: This would also make `invalid_ip`/`invalid_type` cfg-gated which is not necessary

pront: Fair point, those two don’t strictly need the gate today (same above). My thinking is that they’re also fairly lightweight validation cases, so the benefit of keeping them runnable outside the feature gate seems pretty small. I’d prefer the simpler module-level gate because it keeps the DNS test group easier to reason about and avoids future network-dependent cases being added without the intended test feature guard.

pront (approved):
https://github.com/vectordotdev/vrl/pull/1797#discussion_r3377038838

src/stdlib/reverse_dns.rs
thomasqueirozb: The thing is that currently our CI doesn't enable `test`. Something to be improved at a later date for sure.

PR #1800, opened 2026-06-04
chore: update to Rust 1.95 and fix result_large_err warnings

thomasqueirozb:
## Summary

Updates the MSRV to Rust 1.95 and resolves all `clippy::result_large_err` warnings introduced by the updated lint.

- Box the `array` and `object` collection fields inside `Kind` (88 → 24 bytes), eliminating the `result_large_err` warnings that fire when `Kind` appears in `Result` error variants.
- Box the `expr` field in `function::Error::UnexpectedExpression` and `ExpectedStaticExpression` variants for the same reason.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## Is this a breaking change?

- [ ] Yes
- [x] No

## How did you test this PR?

`cargo clippy --lib` passes clean. Existing test suite passes.

**Benchmarks vs `main`:**

`Kind::to_string()` on a pre-built `Kind` (no collection contents):

| case | change |
|------|--------|
| BYTES | −12.8% |
| NULL | −8.3% |
| REGEX | −4.5% |
| ARRAY | +60.4% |

Full realistic error path — `Value::kind()` + format into error message (as in `bool(arr) ?? false`):

| case | change |
|------|--------|
| array, 10 elements | −17.9% |
| object, 10 keys | −7.8% |
| bytes | no change |

The isolated `to_string()` regression on ARRAY (+60%) disappears in the real path because the O(n) iteration over collection elements dominates, and those allocations benefit from the smaller struct. The `parse_xml` stdlib bench (11 cases) showed no significant change (all within ±2%).

- [ ] Yes. Please add a changelog fragment based on our [guidelines](https://github.com/vectordotdev/vector/blob/master/changelog.d/README.md).

## Checklist

- [ ] If this PR introduces changes to [LICENSE-3rdparty.csv](https://github.com/vectordotdev/vrl/blob/main/LICENSE-3rdparty.csv), please run `dd-rust-license-tool write` and commit the changes. More details [here](https://crates.io/crates/dd-rust-license-tool).

## References

NA


src/value/kind/conversion.rs
```diff
@@ -5,48 +5,48 @@ impl Kind {
     ///
     /// This returns `None` if the type is not known to be an object.
     #[must_use]
-    pub const fn as_object(&self) -> Option<&Collection<Field>> {
```
pront, line 8: Nit: we can probably preserve `const` by introducing a match `match self.object.as_ref() `.

pront (approved):

PR #1900, opened 2026-08-21
Add cooperative execution control for interrupting VRL programs

gwenaskell:
## Summary

- Adds an `ExecutionControl` trait and `Runtime::resolve_with_control` letting an embedder cooperatively interrupt a running VRL program between expressions — e.g. after their own deadline elapses, on a client disconnect, or against an operation budget. The existing `Runtime::resolve` is untouched (it's the same code path with no control configured), and there's no Cargo feature: this is always compiled in and a true no-op when unused.
- `Context::checkpoint()` is called once per `Expr::resolve` — the single dispatch point every expression in a program flows through — and returns `Result<(), ExpressionError>`. A controller returning `ControlFlow::Break` turns into `ExpressionError::Interrupted`, which propagates as an ordinary error through VRL's existing `Resolved`/`?` machinery. No panics, no unwind-safety concerns.
- `ExpressionError::Interrupted` / `Terminate::Interrupted` deliberately can't be swallowed by VRL's own error-handling constructs: it's excluded from `??` coalescing, infallible assignment (`value, error = ...`), fallible-function-call (`!`) handling, and boolean `||` short-circuiting. It always reaches the caller of `resolve_with_control`.
- Fixed a real bug along the way: `closure::Runner`'s per-iteration cleanup (restoring a variable shadowed by a closure argument) now always runs before an error propagates, not just on success — previously an early `?` skipped cleanup on *any* error, not only interruption.

## How we got here

This started as a Cargo-feature-gated `AtomicBool` flag checked via a `Context::cancel_breakpoint` that panicked on cancellation. Review turned up two real problems with that design: a panic mid-`for_each` could unwind past closure cleanup and leave stale iteration state on a reused `Runtime`, and a single long-running or blocking function call still couldn't be preempted mid-call. The current `ExecutionControl`/`checkpoint()` design fixes the first properly (see the cleanup fix above, and the `interruption_restores_closure_variables` test) and documents the second as an inherent limitation of checkpointing between expressions rather than during one — see below.

## Known limitation

Checking only happens between expressions, not during one: a function that blocks (e.g. network I/O) or loops internally without itself calling `Context::checkpoint()` still runs to completion before the next checkpoint can fire.

## Test plan

- [x] `cargo test --features default --lib` — 1863 tests pass, including 7 new in `compiler::runtime::execution_control_tests`: existing `resolve` behavior is unaffected without a control, a `for_each` loop is interrupted mid-iteration, interruption isn't caught by `??` / infallible assignment / boolean `||`, a control is scoped to a single `resolve_with_control` call, and closure-shadowed variables are correctly restored after an interruption.
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings` — clean.
- [x] `cargo fmt --check --all` — clean.

### DIFF
--- a/changelog.d/1900.feature.md
+++ b/changelog.d/1900.feature.md
@@ -1,0 +1,7 @@
+Added an opt-in `execution_timeout` Cargo feature that lets embedders bound
+how long a VRL program is allowed to run. When enabled and a timeout is
+configured via `Runtime::set_timeout`, the program panics once wall-clock
+execution time exceeds the deadline. This is a no-op unless the feature is
+enabled and a timeout is set.
+
+authors: gwenaskell
--- a/changelog.d/1900.feature.md
+++ b/changelog.d/1900.feature.md
@@ -1,7 +1,7 @@
-Added an opt-in `execution_timeout` Cargo feature that lets embedders bound
-how long a VRL program is allowed to run. When enabled and a timeout is
-configured via `Runtime::set_timeout`, the program panics once wall-clock
-execution time exceeds the deadline. This is a no-op unless the feature is
-enabled and a timeout is set.
+Added an opt-in `execution_cancellation` Cargo feature that lets embedders
+abort a running VRL program from another thread. When enabled and a flag is
+registered via `Runtime::set_cancellation_flag`, the program checks the flag
+on every expression resolution and panics as soon as it's observed set. This
+is a no-op unless the feature is enabled and a flag is registered.
 
 authors: gwenaskell
--- a/changelog.d/1900.feature.md
+++ b/changelog.d/1900.feature.md
@@ -1,7 +1,0 @@
-Added an opt-in `execution_cancellation` Cargo feature that lets embedders
-abort a running VRL program from another thread. When enabled and a flag is
-registered via `Runtime::set_cancellation_flag`, the program checks the flag
-on every expression resolution and panics as soon as it's observed set. This
-is a no-op unless the feature is enabled and a flag is registered.
-
-authors: gwenaskell
--- a/src/compiler/expression.rs
+++ b/src/compiler/expression.rs
@@ -234,6 +234,8 @@ impl Expression for Expr {
             Return, Unary, Variable,
         };
 
+        ctx.breakpoint();
+
         match self {
             Literal(v) => v.resolve(ctx),
             Container(v) => v.resolve(ctx),
--- a/src/compiler/expression.rs
+++ b/src/compiler/expression.rs
@@ -234,8 +234,7 @@ impl Expression for Expr {
             Return, Unary, Variable,
         };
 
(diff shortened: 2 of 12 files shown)


changelog.d/1900.feature.md
```diff
@@ -0,0 +1,7 @@
+Added an opt-in `execution_cancellation` Cargo feature that lets embedders
+abort a running VRL program from another thread. When enabled and a flag is
+registered via `Runtime::set_cancellation_flag`, the program checks the flag
+on every expression resolution and panics as soon as it's observed set. This
+is a no-op unless the feature is enabled and a flag is registered.
+
+authors: gwenaskell
```
dd-sebastien-lb: I don't know if that should be part of the changelog since this is not a "user facing" modification

src/compiler/expression.rs
```diff
@@ -234,6 +234,8 @@ impl Expression for Expr {
             Return, Unary, Variable,
         };
 
+        ctx.cancel_breakpoint();
```
dd-sebastien-lb: Did you consider making this `#[cfg(feature = "execution_cancellation")]` only to avoid defining an empty func above ?

gwenaskell: good idea, this is actually better to make it explicit it does not make sense using it without the FF

src/compiler/runtime.rs
```diff
@@ -127,3 +131,85 @@ impl Runtime {
...
+        let flag = Arc::new(AtomicBool::new(false));
+        let mut runtime = Runtime::new(RuntimeState::default());
+        runtime.set_cancellation_flag(Arc::clone(&flag));
+
+        let handle = std::thread::spawn(move || {
+            runtime.resolve(&mut target, &program, &TimeZone::default())
+        });
+
+        std::thread::sleep(Duration::from_millis(5));
+        flag.store(true, Ordering::Relaxed);
```
dd-sebastien-lb: Is there a risk of flakyness here ? I guess `500_000` is enough to avoid that

dd-sebastien-lb (approved):

src/compiler/runtime.rs
gwenaskell: good point. This test is actually not super useful, let's remove it

src/compiler/expression.rs
```diff
@@ -234,6 +234,9 @@ impl Expression for Expr {
             Return, Unary, Variable,
         };
 
+        #[cfg(feature = "execution_cancellation")]
+        ctx.cancel_breakpoint();
```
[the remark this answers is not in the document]
gwenaskell: this is a known limitation of this approach.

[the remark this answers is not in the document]
gwenaskell: documented in code

pront (approved):

gwenaskell:
@codex review

PR #1931, opened 2026-09-13
feat(stdlib): add `break` statement support for early loop exit in `for_each`

jimmystewpot:
## Summary

This PR adds native support for the `break` keyword within closure-based iterators in VRL, addressing the long-standing feature request in [vectordotdev/vrl#1480](https://github.com/vectordotdev/vrl/discussions/1480). It's taken some time due to other life issues getting in the way (time since original discussion).

Previously, VRL lacked an early-loop-exit mechanism. Scripts iterating over large arrays or objects with `for_each` to find or process an item had to evaluate all subsequent elements, even after finding a match, using guard flags like `if found == null { ... }`. On large payloads, this incurred significant and unnecessary $O(N)$ runtime overhead.

This change introduces `break` as a language keyword, enforces compile-time semantic checking so it can only appear within breakable loop contexts, and short-circuits `for_each` execution upon encountering a `break`.

### 1. Lexer & Parser

- Promoted `"break"` from `Token::ReservedIdentifier` to `Token::Break` in `src/parser/lex.rs`.
- Added `Break` AST node in `src/parser/ast.rs` and integrated `BreakExpr` into `src/parser/parser.lalrpop`.
- Non-breaking change: `break` was already a reserved identifier in VRL, so no valid user scripts could have used it as a variable name.

### 2. Semantic Analysis & Diagnostics

- Added diagnostic error code `E632: BreakOutsideLoop` (`"break outside of loop or iterator"`).
- Added `in_breakable_context` tracking in `Compiler`. It is incremented when entering closures for iterator functions (`is_iterator == true`).
- Any `break` expression placed outside an iterator closure produces a descriptive compile-time diagnostic error with code 632.

### 3. Runtime & Closure Cleanup

- Added `ExpressionError::Break { span: Span }` to represent loop termination control flow.
- Updated `Runner::run_key_value` and `Runner::run_index_value` in `src/compiler/function/closure.rs` to guarantee that outer shadowed variables are properly cleaned up and restored in the environment before propagating `Break`.
- Updated `for_each` in `src/stdlib/for_each.rs` to intercept `ExpressionError::Break` and immediately terminate the loop, returning `Ok(Value::Null)`.
- Nested loops correctly terminate only the innermost loop.

### 4. Criterion Benchmark (`benches/break_iteration.rs`)

Added a Criterion benchmark suite comparing early search using `break` vs the status-quo pattern of scanning all elements guarded by a boolean flag `found == null` on array sizes of 100, 1,000, and 10,000 elements with the match at index 5:

| Payload Size (`N`) | Early `break` (`O(k)`) | Full Iteration (`O(N)`) | Speedup |
| :--- | :--- | :--- | :--- |
| 100 elements | 1.54 µs | 14.85 µs | ~9.7x faster |
| 1,000 elements | 4.74 µs | 216.85 µs | ~45.8x faster |
| 10,000 elements | 66.75 µs | 3,307.30 µs (3.31 ms) | ~49.5x faster |

## Test Coverage

- Unit Tests: `cargo test --lib` (1,875 passed, 0 failed).
- VRL Test Suite: `cargo run -p vrl-tests` (954 passed, 0 failed).
- Integration Tests Added:
  - `break_outside_loop.vrl`: Diagnostic verification for compile error `E632`.
  - `break_in_for_each_array.vrl`: Verifies array early exit and evaluated state retention.
  - `break_in_for_each_object.vrl`: Verifies object key/value early exit.
  - `break_nested_for_each.vrl`: Verifies that `break` exits only the innermost loop.
- Lints & Formatting: `cargo clippy --all-targets` passes with 0 warnings; `cargo fmt --check` is clean.
- Changelog Fragment: Added `changelog.d/break-in-for-each.feature.md`.

## Related Discussion

- Resolves [vectordotdev/vrl#1480](https://github.com/vectordotdev/vrl/discussions/1480).

### DIFF
--- a/benches/break_iteration.rs
+++ b/benches/break_iteration.rs
@@ -1,0 +1,89 @@
+use criterion::{BenchmarkId, Criterion, criterion_group, criterion_main};
+use vrl::compiler::state::RuntimeState;
+use vrl::compiler::{Context, TimeZone};
+use vrl::value::Value;
+
+const SIZES: [usize; 3] = [100, 1_000, 10_000];
+
+const BREAK_SRC: &str = r#"
+found = null
+for_each(.) -> |_index, item| {
+    if item == 5 {
+        found = item
+        break
+    }
+}
+found
+"#;
+
+const FLAG_SRC: &str = r#"
+found = null
+for_each(.) -> |_index, item| {
+    if found == null {
+        if item == 5 {
+            found = item
+        }
+    }
+}
+found
+"#;
+
+fn benchmark_break_vs_flag(c: &mut Criterion) {
+    let fns = vrl::stdlib::all();
+    let break_prog = vrl::compiler::compile(BREAK_SRC, &fns)
+        .expect("failed to compile break VRL program")
+        .program;
+    let flag_prog = vrl::compiler::compile(FLAG_SRC, &fns)
+        .expect("failed to compile flag VRL program")
+        .program;
+
+    let mut group = c.benchmark_group("for_each_early_exit");
+
+    for &size in &SIZES {
+        let array: Value = (0..size)
+            .map(|i| Value::Integer(i as i64))
+            .collect::<Vec<_>>()
+            .into();
+
+        group.bench_with_input(BenchmarkId::new("break", size), &array, |b, arr| {
+            b.iter_batched(
+                || {
+                    let state = RuntimeState::default();
+                    let target = arr.clone();
+                    (state, target)
+                },
+                |(mut state, mut target)| {
+                    let tz = TimeZone::default();
+                    let mut ctx = Context::new(&mut target, &mut state, &tz);
+                    let res = break_prog.resolve(&mut ctx);
+                    debug_assert_eq!(res, Ok(Value::Integer(5)));
+                    res
+                },
+                criterion::BatchSize::SmallInput,
+            );
+        });
+
+        group.bench_with_input(BenchmarkId::new("flag", size), &array, |b, arr| {
+            b.iter_batched(
+                || {
+                    let state = RuntimeState::default();
+                    let target = arr.clone();
+                    (state, target)
+                },
+                |(mut state, mut target)| {
+                    let tz = TimeZone::default();
+                    let mut ctx = Context::new(&mut target, &mut state, &tz);
+                    let res = flag_prog.resolve(&mut ctx);
+                    debug_assert_eq!(res, Ok(Value::Integer(5)));
+                    res
+                },
+                criterion::BatchSize::SmallInput,
+            );
+        });
+    }
+
+    group.finish();
+}
+
+criterion_group!(benches, benchmark_break_vs_flag);
+criterion_main!(benches);
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -214,7 +214,7 @@ impl<'a> Compiler<'a> {
             Unary(node) => self.compile_unary(node, state).map(Into::into),
             Abort(node) => self.compile_abort(node, state).map(Into::into),
             Return(node) => self.compile_return(node, state).map(Into::into),
-            Break(node) => self.compile_break(node, state).map(Into::into),
+            Break(node) => self.compile_break(&node, state).map(Into::into),
         }?;
 
         // If the compiled expression is fallible and no sub-expression has
@@ -933,7 +933,11 @@ impl<'a> Compiler<'a> {
         })
     }
 
-    fn compile_break(&mut self, node: Node<ast::Break>, _state: &mut TypeState) -> Option<Break> {
+    fn compile_break(
+        &mut self,
+        node: &Node<ast::Break>,
+        _state: &mut TypeState,
+    ) -> Option<Break> {
         let span = node.span();
         if self.in_breakable_context == 0 {
             self.diagnostics.push(Box::new(break_::Error::new(span)));
--- a/src/compiler/compiler.rs
+++ b/src/compiler/compiler.rs
@@ -804,14 +804,10 @@ impl<'a> Compiler<'a> {
                     None => None,
                     Some(block) => {
                         let span = block.span();
-                        let is_breakable = builder.supports_break();
-                        if is_breakable {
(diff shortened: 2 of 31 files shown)


src/compiler/compiler.rs
```diff
@@ -801,7 +804,15 @@ impl<'a> Compiler<'a> {
                     None => None,
                     Some(block) => {
                         let span = block.span();
-                        match self.compile_block_with_type(block, state) {
+                        let is_iterator = builder.is_iterator();
+                        if is_iterator {
+                            self.in_breakable_context = self.in_breakable_context.saturating_add(1);
```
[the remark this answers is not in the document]
jimmystewpot: Resolved in 9966ae93f and f16021f1f.

We replaced the generic `is_iterator()` check with an explicit `supports_break` field on closure definitions (`closure::Definition`), which is set to `true` exclusively for `for_each` and remains `false` for transformation/filtering closures (`filter`, `map_values`, `map_keys`, `replace_with`).

Additionally, closure compilation now strictly scopes breakability per closure level (`self.in_breakable_context = if builder.supports_break() { 1 } else { 0 }; ... self.in_breakable_context = prev;`), preventing closures nested inside a `for_each` from inheriting break capability. We also added compiler diagnostic regression tests for `filter`, `map_values`, and nested closures.

src/compiler/expression_error.rs
```diff
@@ -22,6 +22,9 @@ pub enum ExpressionError {
         span: Span,
         value: Value,
     },
+    Break {
+        span: Span,
+    },
```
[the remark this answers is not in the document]
jimmystewpot: Resolved in 596568875 and aeb727aed.

`ExpressionError::Break` is now treated as unrecoverable control flow across all recovery mechanisms:
1. **Infallible Assignment (`assignment.rs`):** `Variant::Infallible::resolve` now checks `matches!(error, ExpressionError::Interrupted | ExpressionError::Break { .. })` and immediately propagates `Err(error)` instead of populating the `err` variable.
2. **Error Coalescing (`op.rs`):** The `??` operator exempts `ExpressionError::Break` from fallback evaluation and re-raises the error.
3. **Logical OR (`value/error.rs`):** `From<ValueError> for ExpressionError` preserves `ExpressionError::Break` when converting from `ValueError::Or`, preventing `||` from swallowing the break.

TDD unit tests have been added in `compiler::runtime::execution_control_tests` verifying that `break` terminates iteration early across `_, err = ...`, `??`, and `||`.

src/compiler/expression/break_.rs
```diff
@@ -0,0 +1,77 @@
...
+    }
+}
+
+impl Expression for Break {
+    fn resolve(&self, _ctx: &mut Context) -> Resolved {
+        Err(ExpressionError::Break { span: self.span })
+    }
+
+    fn type_info(&self, state: &TypeState) -> TypeInfo {
+        TypeInfo::new(state, TypeDef::never())
```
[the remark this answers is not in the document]
jimmystewpot: Thanks for taking the time to review and for raising this thoughtful observation.

`IfStatement::type_info` currently merges branch states without pruning terminating branches. Looking across the rest of the compiler, this is an existing design characteristic across all terminating control flow in VRL, rather than something specific to `break`. For example, `return` and `abort` behave the same way today—as seen in `lib/tests/tests/expressions/return/return_typedef.vrl` (lines 37–45), where an `if` branch containing `return` still merges its mutated type state into the continuation (`bytes | integer`).

Introducing dead-branch pruning in `IfStatement::type_info` would be a valuable enhancement to make type inference smarter, but because it alters type resolution semantics language-wide for `return`, `abort`, and `break` alike, it would be best addressed in a dedicated follow-up PR so this change remains focused purely on introducing the `break` statement without unintended side-effects.

Thanks for the detailed review.

benches/break_iteration.rs
```diff
@@ -0,0 +1,89 @@
+use criterion::{BenchmarkId, Criterion, criterion_group, criterion_main};
```
pront: Let's remove this file.

jimmystewpot:
@codex review

pront (commented):
Looks good but I need a bit more time to review.

benches/break_iteration.rs
jimmystewpot: Thanks for the review. Removed in 3391ed658 — deleted the benchmark file and cleaned up its entry in Cargo.toml. Please take all the time you need to review.

jimmystewpot:
> Looks good but I need a bit more time to review.

I've moved up some work to sync our fork of vrl's `for` loop implementation and tidy up to create a draft PR #1952 I figured that because there's overlap in 'break' seeing both together might be useful. I've left the other PR in draft because it's 

a) unreasonably large
b) wasn't sure if this should have a pre-approved proposal

src/compiler/compiler.rs
```diff
@@ -62,6 +62,7 @@ pub struct Compiler<'a> {
     pending_fallibilities: Vec<CompilerError>,
 
     config: CompileConfig,
+    in_breakable_context: usize,
```
pront:
```suggestion
    in_breakable_context: bool,
```

src/compiler/expression/assignment.rs
```diff
@@ -540,7 +540,10 @@ where
                     value
                 }
                 Err(error) => {
-                    if matches!(error, ExpressionError::Interrupted) {
+                    if matches!(
```
pront, line 543: IMO this is the only issue with the implementation. This works but now we are using Err to propagate control signals instead of just errors.

This will require a bit of refactoring but we can do:
```rust
enum EvaluationOutcome {
    Value(Value),
    Return(Value),
    Break,
}

type Resolved = Result<EvaluationOutcome, RuntimeError>;
```

But I don't think we need to block this PR. I can follow up with a PR to refactor this.

pront (approved):

hestonhoffman (approved):

src/compiler/compiler.rs
jimmystewpot: I submitted this as a fix (below) and regenerated the documentation (which was failing a check in actions)

PR #1954, opened 2026-09-18
feat(value)!: add Value::String for guaranteed UTF-8

bruceg:
## Summary

The `Value` type has one variant for string-of-bytes values and does not distinguish internally between valid and invalid UTF-8. Since many stored values and many uses of those values requires UTF-8, this results in frequent validation and conversion into the native string types. To reduce this, this change adds a `Value::String` variant with the same semantics as `Bytes` except that it holds known-valid UTF-8 content. This the first step of implementation of rfcs/2026-05-01-value-string-variant.md and is code-level breaking, although VRL programs should not require any changes.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [X] Non-functional (chore, refactoring, docs)
- [X] Performance

## Is this a breaking change?

- [X] Yes
- [ ] No

  our [guidelines](https://github.com/vectordotdev/vrl/blob/main/changelog.d/README.md).
- [X] No. A maintainer will apply the "no-changelog" label to this PR.

## Checklist


changelog.d/value-string-variant.breaking.md
```diff
@@ -0,0 +1,29 @@
+`Value` now has a `String` variant (`bytestring::ByteString`) for guaranteed UTF-8, alongside
+`Bytes` for raw byte sequences. VRL `==` / `Hash` / ordering still compare byte content across the
+two variants. Natural UTF-8 constructors (`From<&str>`, JSON strings, VRL string literals) now
+produce `Value::String`; `From<Bytes>` is unchanged.
+
+This is a compile-time break for exhaustive `match` on `Value`. It is a **silent** skip for
```
pront: Since we do not guarantee compile time backwards compatibility (for now, this might change) we can remove `changelog.d/value-string-variant.breaking.md`.

bruceg: There are some small optimizations in this change, would you like a non-`breaking` changelog entry or leave it as a tech-debt kind of change?

pront: If they are user facing let's add a non-breaking changelog entry

bruceg: On rethinking it, at this point there are no real user-visible changes (user being a VRL author or Vector config writer) so I'll just drop it.

src/value/value.rs
```diff
@@ -34,11 +36,14 @@ pub type StdError = Box<dyn std::error::Error + Send + Sync + 'static>;
...
 
 /// The main value type used in Vector events, and VRL.
-#[derive(Eq, PartialEq, Hash, Debug, Clone)]
+#[derive(Debug, Clone)]
 pub enum Value {
     /// Bytes - usually representing a UTF8 String.
     Bytes(Bytes),
 
+    /// Guaranteed-UTF-8 string.
+    String(ByteString),
```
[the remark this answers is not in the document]
bruceg: "User" in the context are VRL authors which see no changes.

src/value/value/convert.rs
```diff
@@ -115,23 +116,75 @@ impl Value {
...
+    ///
+    /// Prefer [`Self::from_utf8_or_bytes`] when the encoding is not already proven.
+    ///
+    /// # Safety
+    ///
+    /// `bytes` must be valid UTF-8. Passing invalid UTF-8 violates `ByteString`'s
+    /// invariant and is undefined behavior in later `str` operations (deref,
+    /// formatting, serde, display).
+    #[must_use]
+    pub unsafe fn from_utf8_bytes(bytes: Bytes) -> Self {
```
pront:
```suggestion
    pub unsafe fn from_utf8_unchecked(bytes: Bytes) -> Self {
```

src/compiler/expression/literal.rs
```diff
@@ -45,6 +48,19 @@ impl Literal {
             Null => Value::Null,
         }
     }
+
+    /// Build a UTF-8 string literal from bytes that are UTF-8 by construction.
+    ///
+    /// # Safety
+    ///
+    /// `bytes` must be valid UTF-8. See [`Value::from_utf8_bytes`].
+    pub unsafe fn from_utf8_bytes(bytes: Bytes) -> Self {
```
pront:
```suggestion
    pub unsafe fn from_utf8_unchecked(bytes: Bytes) -> Self {
```

src/value/value/convert.rs
```diff
@@ -115,23 +116,75 @@ impl Value {
...
+        match self {
+            Self::Bytes(bytes) => Some(simdutf_bytes_utf8_lossy(bytes.as_ref())),
+            Self::String(s) => Some(Cow::Borrowed(s.as_ref())),
+            _ => None,
+        }
+    }
+
+    /// Returns a `KeyString` if self is `Value::Bytes` or `Value::String`.
+    #[must_use]
+    pub fn to_key_string(&self) -> Option<KeyString> {
```
pront:
```suggestion
    pub fn to_key_string_lossy(&self) -> Option<KeyString> {
```

src/datadog/grok/grok_filter.rs
```diff
@@ -272,14 +255,40 @@ fn scale_value(value: f64, scale_factor: f64) -> Value {
...
+    filter: &GrokFilter,
+    parse: impl Fn(&str) -> V,
+) -> Result<Value, InternalError> {
+    value
+        .as_str()
+        .map(|s| parse(&s).into())
+        .ok_or_else(|| filter_error(filter, value))
+}
+
+fn apply_utf8_filter_error_prone<V: Into<Value>, E: std::error::Error>(
```
pront:
```suggestion
fn try_apply_utf8_filter<V: Into<Value>, E: std::error::Error>(
```

pront (approved):
Let's update function names according to standard library conventions. Otherwise LGTM

bruceg:
Agreed on all counts. Done in cd84166fe

Do you want me to wait to merge until you do the next release point?

src/stdlib/tally.rs
```diff
@@ -3,23 +3,22 @@ use std::collections::{BTreeMap, HashMap};
 
 fn tally(value: Value) -> Resolved {
     let value = value.try_array()?;
-    #[allow(clippy::mutable_key_type)] // false positive due to bytes::Bytes
-    let mut map: HashMap<Bytes, usize> = HashMap::new();
+    let mut map: HashMap<String, usize> = HashMap::new();
```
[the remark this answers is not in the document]
bruceg: I don't think this is valid. The output is an object which has string-only keys, and so we always have to do this conversion either on input or on output. With the previous form, the example two invalid keys would still turn into the same output key, just one of the counts would overwrite the other instead of summing up. Since the sum behavior seems more natural, I think we want to keep this change.

pront:
> Agreed on all counts. Done in [cd84166](https://github.com/vectordotdev/vrl/commit/cd84166fe109b6171e2261b28321fc17192e76dc)
> 
> Do you want me to wait to merge until you do the next release point?

@bruceg no need to wait for a future date. And since all Codex threads are resolved we can queue anytime now.

bruceg:
> > Do you want me to wait to merge until you do the next release point?
> 
> @bruceg no need to wait for a future date. And since all Codex threads are resolved we can queue anytime now.

We discussed this off of GitHub and decided to wait for the next VRL release next week before merging this in to give it a little more time to soak before being released.

